PR #5, opened 2015-06-24
Live fastFT feed

charypar:
Renders fastFT with React server-side and client-side. Also adds a json endpoint for fastFT the client can poll and update the feed. Eventually, we'll switch from polling to listening on the fastFT pusher channel.

## Todo

- [x] Render client-side
- [x] Render server-side
- [x] fastFT JSON endpoint
- [x] Poll and re-draw
- [x] sort out `o-date` integration
- [x] style items and hover state


charypar:
I turned off the linting as part of test for now, because it breaks on JSX, we'd need to enable that in next build tools. I'd like to have a wider opinion on using it before doing that.

imranolas:
:+1:

PR #6, opened 2015-06-24
Layout

imranolas:
- [x] Implement new layout across all breakpoints
- [x] Genre colours


charypar:
Looks good, but could use a bit of CSS output size optimisation, I think there's quite a bit of repetition going on with topic colours and stuff...

client/components/story-card/main.scss
```diff
@@ -1,22 +1,39 @@
...
+
+.story-card {
+	background-color: oColorsGetPaletteColor('warm-base');
+	float: left;
 	display: block;
 	overflow: hidden;
 	padding-bottom: 42px; // TODO Footer height variable
+	position: relative;
+
+	@include nhGridRespondTo($until: XS) { margin-bottom: 10px; }
```
kaelig: This could be simplified with `margin-bottom: 10px;` (no need for a media query here)

< /nosy >

client/components/story-card/main.scss
```diff
@@ -79,14 +76,44 @@
 }
 
 @mixin nhLargeCard {
+	@include nhGridRespondTo($until: M) {
```
kaelig: Ditto.

client/components/story-card/main.scss
charypar: oh yea, will fix.

client/components/grid/main.scss
```diff
@@ -32,8 +32,7 @@
 @mixin nhGridColmargins($total-cols: $nh-grid-columns) {
 	$subgrid-ratio: $nh-grid-columns / $total-cols;
 
-	margin-left: percentage($subgrid-ratio * $nh-grid-gutter / 2);
-	margin-right: percentage($subgrid-ratio * $nh-grid-gutter / 2);
+	margin: percentage($subgrid-ratio * $nh-grid-gutter / 2);
```
charypar: Shall we rename the mixin to just GridMargins then?

client/components/grid/main.scss
```diff
@@ -123,11 +122,17 @@
...
 	}
-	@include nhGridRespondTo($from: XL) {
-		width: $_nh-grid-max-width;
+	@include nhGridRespondTo($from: S, $until: M) {
+		width: 730px;
+	}
+	@include nhGridRespondTo($from: M, $until: L) {
+		width: 970px;
+	}
+	@include nhGridRespondTo($from: L) {
```
charypar: Can we extract these into another configuration map in `_variables` as a `!default` so people can override the container widths?

client/components/story-card/main.scss
```diff
@@ -1,22 +1,39 @@
...
+	.story-card__footer { border-bottom: 1px $color solid; }
+	.story-card__taxon { color: $color; }
+}
+
+.story-card {
+	background-color: oColorsGetPaletteColor('warm-base');
+	float: left;
 	display: block;
 	overflow: hidden;
 	padding-bottom: 42px; // TODO Footer height variable
```
charypar: Worth adding the variable before merging? :)

client/components/story-card/main.scss
```diff
@@ -30,39 +47,17 @@
...
+	&.lifestyle { @include nhStoryGenreColors(map-get($o-colors-palette, 'lifestyle')); }
+}
 
-	@include nhGridRespondTo(XS) {
-		.story-card__taxon { font-size: 20px; }
-		.story-card__title { font-size: 28px; }
-		.story-card__summary { font-size: 24px; }
-	}
+@mixin nhStoryCard($type) {
+	// &.lifestyle { @include nhStoryGenreColors(map-get($o-colors-palette, 'lifestyle')); }
```
charypar: remove?

client/components/story-card/main.scss
```diff
@@ -79,14 +76,42 @@
...
 		display: block;
 	}
 
-	.story-card__summary { display: initial; }
+	.story-card__content { @include nhGridColumn(3, 9); }
 
+	.story-card__image,
+	.story-card__content {
+		@include nhGridRespondTo($until: XS) {
+			@include nhGridColumn(9, 9);
```
charypar: Is this assuming story card lives in a 9 column grid? If so, should we make the total number of columns an optional argument to the top level card mixin?

client/components/story-card/main.scss
```diff
@@ -94,39 +119,62 @@
...
 
 @mixin nhWideCard {
+	font-size: 24px;
+	@include nhGridRespondTo($from: M) {
+		.story-card__title { font-size: 1em; }
+		// .story-card__taxon { font-size: 0.8em; }
+	}
 	.story-card__image {
-		width: initial;
+		@include nhGridColumn(4, 9);
```
charypar: Same here

server/init.js
```diff
@@ -5,7 +5,8 @@ import frontPage from './routes/front-page';
 
 var app = express({
 	helpers: {
-		dump: (it) => JSON.stringify(it, null, 2)
+		dump: (it) => JSON.stringify(it, null, 2),
```
charypar: This actually already exists and it's called `json`, I removed it in mine. (See <https://github.com/Financial-Times/next-handlebars>)

views/partials/lead-today.html
```diff
@@ -0,0 +1,19 @@
+<div class="flexipod flexipod--lead-today">
```
charypar: Is there a good way to structure the `views` folder so that it's clear which component the templates belong to? Or even move the templates into those components maybe...?

client/components/story-card/main.scss
imranolas: No, the story cards are themselves a 9 column grid. This make no assumptions about where the card is used.

PR #13, opened 2015-06-29
Use next-sass-setup rather than its indidvidual dependencies (resolves #12)

### LINKED ISSUE #12: Switch back to next-sass-setup
This requires a bit of work on `next-sass-setup` and updates to how we use `o-colors`.

charypar:
- [x] include `next-sass-setup`
- [x] use `o-colors` properly

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -7,7 +7,6 @@
     "o-date": "~1.1.1"
   },
   "resolutions": {
-    "o-colors": "^3.0.0",
     "o-fonts": "^2.0.0",
     "o-ft-icons": "^2.3.4"
   }
--- a/bower.json
+++ b/bower.json
@@ -2,9 +2,7 @@
   "name": "next-homepage",
   "dependencies": {
     "next-express": "^10.0.0",
-    "o-colors": "~2.5.0",
-    "next-type": "~1.3.3",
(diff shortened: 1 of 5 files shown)


charypar:
I also realised we do need o-grid generated anyway for header and footer, so there is no need to change next-sass-setup.

charypar:
@keirog, @kaelig would you mind having a look at this before we merge to make sure we're not doing something stupid?

keirog:
Looks fine to me. General comment on color usage though: I imagine the Origami folks would be keener for you to use one of their [o-colors mixins/functions](https://github.com/Financial-Times/o-colors#use-case-mixin), rather than a `map-get` on the the palette map… @kaelig?

charypar:
@keirog I looked for one to define new custom colors, but couldn't find one, hence extending the palette with `map-merge`. It's meant to be temporary until we settle on a set of colors and are happy, then merging it int `o-colors` and defining use-cases properly.

charypar:
Hang on, misread your comment, are we using `map-get` anywhere?

charypar:
@keirog using `oColorsGetPaletteColor` now.

keirog:
:+1: nice, go for it

PR #15, opened 2015-06-29
Editors picks

imranolas:
- [x] Build editors picks from multiple sources
- [x] Add front-page route for prod testing
- [x] Add colors
- [x] Build editors picks module


client/main.scss
```diff
@@ -30,67 +32,93 @@ html {
...
+
+
 body {
 	@include nhGridGenerate();
 	padding: 0;
 	margin: 0;
 }
 
+.fixed-width-container {
+	@each $layout in $nh-container-widths {
```
charypar: can we sink this into the nh-grid as a default variable? A map with the widths seems appropriate.

charypar: Or you know what, ignore me. We don't really want that in the long term, so lets not put it in there...

PR #38, opened 2015-06-30
Mobile fastFT header

charypar:
Adds a direct link to fastFT stream on smaller screens as a stopgap solution before we have tab control toggle.


imranolas:
:+1:

PR #43, opened 2015-06-30
Genre colors

### LINKED ISSUE #26: Missing genre colour
There's a story that doesn't have a genre colour, we need to find out what tag it has and give it a colour!

screen shot 2015-06-30 at 10 40 03

imranolas:
- [x] next-article-genres for handling all genre colours (fix #26 )


client/components/story-card/main.scss
```diff
@@ -6,12 +6,10 @@
 	.story-card__taxon, .carousel-card__taxon { color: $color; }
 }
 
-.news { @include nhStoryGenreColors(map-get($o-colors-palette, 'news')); }
-.comment { @include nhStoryGenreColors(map-get($o-colors-palette, 'opinion')); }
-.lifestyle,
-.review { @include nhStoryGenreColors(map-get($o-colors-palette, 'lifestyle')); }
-.analysis { @include nhStoryGenreColors(map-get($o-colors-palette, 'analysis')); }
-.features { @include nhStoryGenreColors(map-get($o-colors-palette, 'features')); }
+.default { @include nhStoryGenreColors(map-get($o-colors-palette, 'claret')); }
```
charypar: can you please use `oColorsGetPaletteColor`?

server/init.js
```diff
@@ -16,7 +17,8 @@ var app = express({
 				return img.type === type;
 			});
 			return image ? `https://next-geebee.ft.com/image/v1/images/raw/${image.url}?source=next&fit=scale-down&width=710` : null;
-		}
+		},
+		getGenre(metadata) { if (metadata) { return articleGenres(metadata); } }
```
charypar: Should we rather do this once as a post-processing step on the fetched data?

client/components/story-card/main.scss
kaelig: Also, you might want to prefix these class names to avoid style clashes (especially with `.default` and `.comment`, these are very common names).

PR #50, opened 2015-07-06
[WIP] Popular feed

imranolas:
- [x] Fetch popular feed
- [x] Display popular feed

### DIFF
--- a/components/feed/fastft/main.scss
+++ b/components/feed/fastft/main.scss
@@ -5,7 +5,6 @@
 	a {
 		display: block;
 		width: 100%;
-		hight: 100%;
 	}
 
 	.fastft__logo__fast {
--- a/components/feed/fastft/main.scss
+++ b/components/feed/fastft/main.scss
@@ -1,0 +1,29 @@
+@import "next-type/main";
+@import "o-colors/main";
+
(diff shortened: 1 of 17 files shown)


components/feed/fastft/main.scss
```diff
@@ -0,0 +1,29 @@
+@import "next-type/main";
+@import "o-colors/main";
+
+@mixin nhFastFtTitle {
+	a {
+		display: block;
+		width: 100%;
+		hight: 100%;
```
charypar: height

server/lib/content.js
```diff
@@ -92,15 +94,24 @@ const cachedContent = (topStoriesRegion) => {
...
+						publishedDate: it.item.publishedDate || it.item.lifecycle.lastPublishDateTime
+					};
+				});
+				obj[feed] = {items: result};
+			return obj;
+			}
+		}, {});
 
-		return Object.assign({}, contentCache[src], {top: top, fastFt: fastFt});
+		return Object.assign({}, contentCache[src], {top: top}, feeds);
```
charypar: can we use the `...` operator instead of `Object.assign`?

PR #52, opened 2015-07-06
Use next-article-primary-tag for displaying topics

charypar:
Columnist's names should now show up.


server/routes/front-page.js
```diff
@@ -5,6 +5,8 @@ module.exports = function(req, res) {
 	const useElasticSearch = res.locals.flags.elasticSearchItemGet.isSwitchedOn;
 	const contentData = content.uk(useElasticSearch);
 
+	console.log(contentData.top.items);
```
imranolas: Log here ;)

PR #60, opened 2015-07-07
Display popular feed on mobile

imranolas:


client/main.scss
```diff
@@ -132,13 +132,16 @@ body {
 	@include nhGridRespondTo($from: S) { margin-bottom: 20px; }
 }
 
-body .sidebar {
-	display: none;
+.sidebar {
+	width: 100%;
```
charypar: You can do this in the markup with `data-nh-grid-colspan="12 S4 M4"`

PR #66, opened 2015-07-07
Editors picks images

imranolas:
Display images for all breakpoints as per design


client/components/carousel-card/main.scss
```diff
@@ -2,10 +2,15 @@
 
 .carousel-card {
 	width: 100%;
-	text-align: center;
+	text-align: left;
 
 	@include nhGridRespondTo(XS) {
+		@include nhGridColspan(6, 12);
```
charypar: `@include nhGridColumn(6)`

client/components/carousel-card/main.scss
```diff
@@ -34,12 +39,25 @@
 	}
 }
 
+.carousel-card__content {
```
charypar: I think this entire rule could be replaced with `@include nhGridColumn((default: 4, S: 12))`

views/partials/editors.html
```diff
@@ -1,7 +1,7 @@
 <div class="flexipod flexipod--editors">
 	<h1 class='carousel-heading'>{{title}}</h1>
 	<div>
-		<div class="article-group">
+		<div class="article-group nh-grid-row">
```
charypar: shouldn't be needed any more

PR #75, opened 2015-07-08
[Ready] GraphQL!

charypar:
Trying out GraphQL for content fetching. It's working really nicely so far, feels like a great way of centralising the content fetching while allowing some post processing that's only needed for next as a product (e.g. brand trumps genre as a primary tag).

This will likely stay here for a while, as I'm only playing with it in the evenings.

# To Do

- [x] lead pod
- [x] fastFT
- [x] related articles
- [x] editors picks
- [x] opinions
- [x] topics
- [x] popular
- [x] surface a GraphQL endpoint (_security...?_) ~~and fetch live fastFT from it~~ separate PR
- [x] separate backend fetching
- [x] caching + polling
- [x] move query/queries to configuration and clearly separate fetching from presentation
- [x] support `useElasticSearch` flag


server/graphql/content-types.js
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+});
+
+const Image = new GraphQLObjectType({
+	name: "Image",
+	description: "An image",
+	fields: () => ({
+		src: {
+			type: GraphQLString,
```
albertstill: Custom scalar of URL would be cool here that validates it's a URL. Have you seen an example with graphql-js? I can't find one in the specs.

server/graphql/content-types.js
```diff
@@ -0,0 +1,135 @@
...
+		},
+		items: {
+			type: new GraphQLList(Content),
+			description: "Content items of the page",
+			resolve: (page) => {
+				return ApiClient.contentLegacy({
+					uuid: page.items,
+					useElasticSearch: true
+				})
+				.then(articles => {
```
albertstill: you can just return a promise here graphql-js checks if it's thenable

server/graphql/content-types.js
HurricaneJames: I see the attraction, but supporting custom scalar types that check the contents of the data, as opposed to just the type, feels like a bad idea. It is probably better to handle validation in the business logic (server and/or client). Of course I see the attraction of a url scalar type and could be persuaded to go either way with enough logic.

server/graphql/content-types.js
charypar: good spot, that's left over from some debug output :)

server/graphql/content-types.js
charypar: I tried a custom scalar for Date and didn't have much luck with it. Need to explore further. @HurricaneJames is right about validation being an app concern. In case of URL there is probably not much point doing more than a string, unless you want to do complete parsing in GraphQL (i.e. having it be an Object type which has `protocol`, `host`, ...).

albertstill: Hmm yeh I see your point. They talk of a custom Time & URL scalar in the specs tho http://facebook.github.io/graphql/#sec-Scalars. I'd be keen to see a implementation with GraphQL.js.

charypar: Guess that's a fair point. I'll look into building those types in later on.

server/graphql/content-types.js
```diff
@@ -0,0 +1,354 @@
...
+const Collection = new GraphQLInterfaceType({
+	name: "Collection",
+	description: "Set of items of type Content",
+	fields: () => ({
+		title: { type: GraphQLString },
+		url: { type: GraphQLString },
+		items: {
+			type: new GraphQLList(Content),
+			args: {
+				from: { name: 'from', type: GraphQLInt },
```
albertstill: `name` declarations are redundant in `args`, it gets set to the value of the key. https://github.com/graphql/graphql-js/pull/37

charypar: Oh nice, thanks.

commuterjoy:
i worry this solves a problem we didn't have.

charypar:
@commuterjoy we certainly felt it.

server/graphql/backend.js
```diff
@@ -0,0 +1,157 @@
...
+		.then((it) => {
+			let expire = now + ttl;
+
+			this.contentCache[key] = {
+				expire: expire,
+				data: it
+			}
+
+			return it;
+		}.bind(this));
```
imranolas: Not sure this is needed. The arrow function should lexically bind `this` for you.

charypar: Oh, TIL.

commuterjoy:
@charypar explain in plain english?

charypar:
@commuterjoy well it cleans up a couple things:

1) All the nested backend queries get planned and executed for us and we get things like full related content without needing to worry about post-processing it - just state related content is fetched like this and it's a `Content`
2) The API now produces geebee image urls on request, instead of doing it with a helper in the view
3) We no longer need a special presentation for the FastFT json endpoint to strip down the payload (the json endpoint will get removed in favour of client supplied query).
4) All the different post-processing business logic (genres, primary tags) is now in the schema and done once for each type. And it solves the problem of different output formats of v1 and v2.

It's essentially a data fetching and post processing layer we were missing and things were starting to get messy, which made us compromise on things like the related content for a lead story, which was now really easy to do properly.

commuterjoy:
Thanks.

> which made us compromise on things like the related content for a lead story

What was the compromise? I've found elasticsearch to be quite flexible.

I'm interested in the limitations of our approach.

charypar:
The compromise was around related stories for the lead story. We ended up only reading data from the `packaging`, because fetching the stories would make the whole thing more complex with another layer of fetching and data merging and processing. It's of course not impossible to handle, just slowly gets out of hand and didn't seem worth it. With GraphQL that just came basically for free.

It's not a total limitation of consuming REST directly, you can of course do all of it - it's just that the deeper into the relationship graph you get the less manageable it gets and the more manual work you need to do to fetch and process the associated content, even though it's content you already know how to process. There are also slight differences in things you need for related content and for top level content, so you can straight up reuse the fetching code. 

GraphQL just takes a lot of it away - once the types are defined, it's quite easy to nest things and fetch exactly what you need just by adding a field and changing the query - everything nested under that field is already done. And the benefit is anyone can ask for exactly what they need without the server needing to change at all (as long as the things exist obviously).

If you refactored a complex enough REST fetching layer for long enough you'd probably end up with something very similar anyway.

commuterjoy:
> With GraphQL that just came basically for free.

Yes, so automatic of expansion of REST resources is a helpful thing.

PR #88, opened 2015-07-08
Fixes

### LINKED ISSUE #80: Dotted dividers between Life & Arts and Markets, and Markets and Technology

### LINKED ISSUE #81: Make sure editors picks don't contain stories displayed elsewhere.
Engels handles this correctly.

imranolas:
- [x] Dotted borders between topics (fix #80)
- [x] Set line-heights to 1.2
- [x] Remove lead article from editors picks. (fix #81)

### DIFF
--- a/client/components/story-card/main.scss
+++ b/client/components/story-card/main.scss
@@ -46,7 +46,7 @@
 .story-card__related-link {
 	@include nTypeCharlie(3);
 	font-size: 1em;
-	line-height: 1.2em;
+	line-height: 1.2;
 	margin-bottom: 10px;
 	font-style: normal;
 }
@@ -72,7 +72,8 @@
 }
 .story-card__summary {
 	@include nTypeDelta(3);
-	line-height: 1.2em;
+	line-height: 1.2;
+
 	font-size: 18px;
 }
 .story-card__content {
--- a/client/components/story-card/main.scss
+++ b/client/components/story-card/main.scss
@@ -46,6 +46,8 @@
 .story-card__related-link {
 	@include nTypeCharlie(3);
 	font-size: 1em;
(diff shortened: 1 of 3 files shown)


client/components/story-card/main.scss
```diff
@@ -70,6 +72,7 @@
 }
 .story-card__summary {
 	@include nTypeDelta(3);
+	line-height: 1.2em;
 	font-size: 18px;
```
i-like-robots: Is there a reason to mix font size and line height units? Why specify a line height unit at all?

imranolas: `nTypeDelta` sets both the font size and line-height in px which in this case needs to be overridden to fulfil the design requirements. As for the mixed units, it feels more manageable to keep line height as a relative measure proportional to font size, that way we can change the font size without  breaking the line height.

kaelig: In that case I'd recommend a unitless line-height so that children keep the same ratio.

imranolas: Good idea :+1:

kaelig: Friendly note: when the two projects merge (homepage + rest of Next), think about adding new sizes directly into next-type if absolutely needed, as every exception to the rule hinders both design consistency and maintainability.

imranolas: I absolutely agree and that is very much our intention, however, at the moment the designs are in flux and not worth bringing into the rest of Next until they settle down a bit.

kaelig: Great to hear. Totally understand and support your approach.

charypar:
#81 also applies to opinions and topics, this is just a partial fix...

PR #100, opened 2015-07-09
TRACK ALL THE THINGS

charypar:
Added data-trackable to all links and most logical page sections. Hopefully this should give us reasonable usage data.

@commuterjoy does it look ok?


commuterjoy:
+1

PR #105, opened 2015-07-09
Brands should link to brand page

### LINKED ISSUE #91: Tax tags of individual writers links to overall "Columnists" stream page
When a user clicks on a name, it should go to the article stream page for that specific author, so that readers can check out all the articles by that author.

imranolas:
fix #91

### DIFF
--- a/views/partials/carousel-card.html
+++ b/views/partials/carousel-card.html
@@ -10,7 +10,7 @@
       </a>
     {{/if}}
     <div class="carousel-card__content">
-      <a href="/stream/sectionsId/{{item.metadata.primarySection.term.id}}" class="story-card--link" tabindex="0" >
+      <a href="/stream/sectionsId/{{primaryTag.id}}" class="story-card--link" tabindex="0" >
         <h3 class="story-card__taxon">{{primaryTag.name}}</h3>
       </a>
     	<a href="/{{item.id}}" class="carousel-card--link">
--- a/views/partials/story-card.html
+++ b/views/partials/story-card.html
@@ -10,7 +10,7 @@
 		</a>
 	{{/if}}
 	<div class="story-card__content">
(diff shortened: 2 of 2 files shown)


views/partials/story-card.html
```diff
@@ -10,7 +10,7 @@
 		</a>
 	{{/if}}
 	<div class="story-card__content">
-		<a href="/stream/sectionsId/{{item.metadata.primarySection.term.id}}" data-trackable="primary-tag" class="story-card--link" tabindex="0" >
+		<a href="/stream/sectionsId/{{primaryTag.id}}" data-trackable="primary-tag" class="story-card--link" tabindex="0" >
```
matthew-andrews: i think this might need to be `{{primaryTag.taxonomy}}/{{primaryTag.id}}`

views/partials/carousel-card.html
```diff
@@ -10,7 +10,7 @@
       </a>
     {{/if}}
     <div class="carousel-card__content">
-      <a href="/stream/sectionsId/{{item.metadata.primarySection.term.id}}" class="story-card--link" tabindex="0" >
+      <a href="/stream/sectionsId/{{primaryTag.id}}" class="story-card--link" tabindex="0" >
```
matthew-andrews: again suspect this might need to be `href="/stream/ {{primaryTag.taxonomy}}/{{primaryTag.id}}"`

views/partials/story-card.html
imranolas: Yes, good spot. In fact, it seems to need an `Id` suffix too,  `href="/stream/{{primaryTag.taxonomy}}Id/{{primaryTag.id}}"`

matthew-andrews: oops yup

PR #106, opened 2015-07-09
Fix broken brand/sections links

### LINKED ISSUE #107: Tax link are broken for most tags
Links contain `/sectionsId/` but for most links that segment is something else, `regionsId` for instance.

imranolas:
fix #107


matthew-andrews:
:+1:

matthew-andrews:
looks like a bower install problem?

PR #110, opened 2015-07-10
Accessibility

charypar:
This should massively improve assistive devices usage on the front page. The markup is tested with VoiceOver on OS X

# To Do
- [x] Proper labels and read order (make things make sense)
- [x] Good keyboard navigation
- [x] fastFT live updates

### DIFF
--- a/client/main.scss
+++ b/client/main.scss
@@ -45,6 +45,15 @@ img {
 	border: none;
 }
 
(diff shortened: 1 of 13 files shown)


charypar:
This is good to go. Wish we had some actual visually impaired users...

PR #118, opened 2015-07-16
Add myFT Tray

adgad:
/cc @charypar @imranolas @wheresrhys 

myft-tray-hp

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,8 @@
     "next-sass-setup": "^6.2.6",
     "sass-mq": "^3.1.2",
     "o-date": "^1.1.1",
-    "next-myft-ui": "^3.0.0"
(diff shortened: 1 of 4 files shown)


charypar:
Also need to discuss with Bede. Don't merge.

commuterjoy:
+1 good collaboration between teams here.

adgad:
Yup, I think James is on top of speaking to Bede, will wait till then :)

commuterjoy:
@charypar 

> +{{#if @root.flags.myFTTray}}

it's behind a flag

charypar:
As per email from James Webb - This won’t be implemented on the homepage for now. It was agreed that recency and frequency team would work with us to figure out an alternative design approach for the red blob once we complete work on the navigation.

PR #128, opened 2015-07-23
switch to circle ci

matthew-andrews:


.travis.yml
```diff
@@ -1,12 +0,0 @@
-sudo: false
-script:
-- make install && make test && make provision
```
charypar: how come these are not needed anymore?

matthew-andrews: centralised http://git.svc.ft.com:8080/projects/NEXTPRIVATE/repos/config-vars/browse/models/continuous-integration.js#40

matthew-andrews: (Sadly not possible to do for all configuration :disappointed: )

PR #130, opened 2015-07-23
Sauce credentials set by config-vars

matthew-andrews:


matthew-andrews:
Tests failing because no test file, the article on is a pretty good base:-
https://github.com/Financial-Times/next-article/blob/master/test/browser/tests/jssuccesstest.js

matthew-andrews:
replaced by #153

PR #133, opened 2015-07-29
Enable docker for production deploys

matthew-andrews:
I know you wanted to do this but I want to move away from the old deployment mechanism asap cc @charypar.

Local dev with Docker not yet supported but shouldn't be hard… if you fancy having a crack at that come chat with me and we'll make a plan :).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,13 +45,13 @@ provision:
 	nbt provision ${TEST_HOST}
 	nbt configure ft-next-front-page ${TEST_HOST} --overrides "NODE_ENV=branch"
 	nbt deploy-hashed-assets
-	nbt deploy ${TEST_HOST} --skip-enable-preboot
+	nbt deploy ${TEST_HOST} --skip-enable-preboot --docker
 	make smoke
 
 deploy:
 	nbt configure
 	nbt deploy-hashed-assets
-	nbt deploy
+	nbt deploy --docker
 
 clean-deploy: clean install build-production deploy
 


charypar:
That's alright, thanks for doing it. I've been busy with o-grid and some other stuff and I'm trying to limit the context switching as much as I can. I'll have a look at the dev setup at some point and come talk to you.

In the future would you please wait for me or @imranolas (or anyone else who focuses on the front-page work in the future) to +1 or merge PRs? I know this one is tiny and safe and next as a whole is following a more relaxed workflow, but it's a rule we're trying to follow to make sure everything has at least two pairs of eyes on it before it goes to prod (it's easier than deciding every time whether that particular PR is worth a code review). Thanks!

matthew-andrews:
Just spotted this after committing [this](https://github.com/Financial-Times/next-front-page/commit/4c64e3eb32b321a93844c7e2aa02879abfd4edfc) to master.  (Without that [this](https://github.com/Financial-Times/next-build-tools/pull/161/files) would have broken your builds)

Will bear this in mind in future.

matthew-andrews:
FYI `nbt rebuild` will currently clear Circle CI caches and rebuild master branches of all apps.  This will bring in any updated dependencies that have been updated centrally for all apps.

If you want to be in control of all changes before they go to prod I strongly suggest using exact version numbers for bower.json and package.json (or shrinkwrapping for package.json).  And I can add an exception for the front page.

PR #136, opened 2015-07-29
Disable retries when requests fail

matthew-andrews:


charypar:
I'm not entirely convinced this is a good way to go. The retries can help with short-term issues (which can be pretty critical after start or restart when we have no content cached) and are otherwise harmless, aren't they?

matthew-andrews:
ok let's not bother

PR #140, opened 2015-07-31
FastFt feed notifications

imranolas:
Instead of polling the content API directly for FastFt updates, we can now poll the notifications API and use the response to determine whether or not any pending articles are waiting to be fetched. This should be a little friendlier to the content APIs.

### DIFF
--- a/server/graphql/fastFtFeed.js
+++ b/server/graphql/fastFtFeed.js
@@ -1,0 +1,69 @@
+import fetch from 'isomorphic-fetch';
+import {fastFt as config} from './config/sources.js';
+import ApiClient from 'next-ft-api-client';
+
+// Polls for changes on the notification api to determine whether a fetch should
+// be made for new content. Hopefully this is a little nicer to the content api
+// polling it directly.
+
+class FastFtFeed {
+	constructor(elasticSearch) {
+		this.elasticSearch = elasticSearch;
+		this.type = (elasticSearch ? 'elasticsearch' : 'capi');
+
+		// in-memory content cache
+		this.contentCache = {};
+		this.since = new Date().toISOString();
+		this.fetchFastFt();
+		this.pollUpdates();
+	}
+
+	fetchFastFt() {
+		const {uuid} = config;
+		return ApiClient.contentAnnotatedBy({
+			uuid: uuid,
+			useElasticSearch: this.elasticSearch
+		})
+		.then(ids => {
+			this.contentCache = {
+				title: "fastFt",
+				conceptId: uuid,
+				sectionId: null,
+				items: ids.slice()
+			};
+			return this.contentCache;
+		}).catch(console.error);
+	}
+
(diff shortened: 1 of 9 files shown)


server/graphql/fastFtFeed.js
```diff
@@ -0,0 +1,67 @@
...
+					this.since = new Date().toISOString();
+				}
+			})
+			.catch(console.error);
+		}, 25 * 1000);
+	}
+	// Requests a list of notifications for FastFT to determine whether there are
+	// any new items.
+	hasNewUpdates() {
+		const url = `http://api.ft.com/content/notifications?since=${this.since}&apiKey=gzaxydnumy4y7gtvkkjbtcbv`;
```
commuterjoy: do we want api keys in the public code base?

charypar: Nope, we don't.

charypar: And I guess this means we need a new one now...

commuterjoy: Yes please :)

Do you need a hand?

imranolas: Doh! Fetching the key from env now. We will need a new one.

charypar: And we should burn this one.

charypar:
@imranolas is this ready, or still need changing API key before we ship?

imranolas:
@charypar: This is ready to go. The key has been destroyed, regenerated and added to `config_vars`.

server/graphql/fastFtFeed.js
```diff
@@ -0,0 +1,67 @@
...
+		return fetch(url)
+			.then(res => res.json())
+			.then(json => !!json.notifications.length)
+			.catch(console.error);
+	}
+
+	fetch() {	return this.contentCache; }
+}
+
+// expire old content after 10 minutes
```
charypar: That comment is no longer relevant

server/lib/graphql.js
```diff
@@ -2,12 +2,16 @@ import {graphql} from 'graphql';
...
 
-const fetch = (backend) => {
+const fetch = (useElasticSearch) => {
 	return (query) => {
 		const then = new Date().getTime();
 
-		return graphql(schema, query, {backend: backend})
+		return graphql(schema, query, {
+			backend: backend(useElasticSearch),
+			fastFtFeed: fastFtFeed(useElasticSearch)
```
charypar: If we're doing this, should we split out the other "different" sources? Like Popular or liveblog updates? Or is that an implementation detail that shouldn't leak into the schema, as the goal is it'll eventually all get moved into CAPIv2...?

charypar:
Ok, I'm gonna merge this and do the refactor in the liveblog branch, which has already become pretty messy and needs the same thing really. Hopefully it'll make much more sense once done

PR #149, opened 2015-08-04
[WIP] Webpack

imranolas:
A webpack config for building the JS bundle. The build step is much faster than `nbt` and when running `--watch` the updates are applied almost instantly. All in all, a much nicer developer experience.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ build-production:
 	nbt about
 
 watch:
-	nbt build --dev --watch
+	nbt build --dev --skip-js --watch & webpack --watch
 
 clean:
 	git clean -fxd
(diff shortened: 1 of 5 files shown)


charypar:
Can we also run `webpack --watch` as part of `make watch`? Guessing it'll need some kind of output multiplexing...

charypar:
@matthew-andrews this may be interesting as part of `nbt`, we can try for a while and report back the results. The JS build is definitely much faster.

matthew-andrews:
Nice work.  I'm quite keen for this to go into OBT rather than NBT.  They were looking into this.  Will post a message in origami Slack.

PR #187, opened 2015-08-12
fix #137: fastFT feed not updating

### LINKED ISSUE #137: FastFT not live rendering updated content
The right content gets fetched from the backend, but not rendered into the page.

imranolas:
fixes #137

Also adds visual feedback (fade) for newly added feed items

### DIFF
--- a/server/routes/fastft.js
+++ b/server/routes/fastft.js
@@ -9,9 +9,10 @@ module.exports = function(req, res) {
 		'Cache-Control': 'max-age=20, public' // 20 seconds
 	});
 
-	graphql(useElasticSearch, mockBackend).fetch(queries.fastFT)
-	.then(data => {
-		res.json(data.fastFT);
-	})
-	.catch(console.log);
+	graphql(useElasticSearch, mockBackend)
+		.fetch(queries.fastFT)
+		.then(data => {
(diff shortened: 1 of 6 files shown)


server/routes/fastft.js
```diff
@@ -6,7 +6,8 @@ module.exports = function(req, res) {
 	const mockBackend = res.locals.flags.mockFrontPage;
 
 	res.set({
-		'Cache-Control': 'max-age=20, public' // 20 seconds
+		'Cache-Control': 'max-age=20, public', // 20 seconds
+		'Surrogate-Control': 'max-age=20, public' // Fastly cache
```
charypar: I'm not quite sure how this works with Fastly. Should we remove `Cache-Control` at the same time? Can you help, @commuterjoy?

charypar: Based on this https://docs.fastly.com/guides/tutorials/cache-control-tutorial#cache-control:-max-age I think the change as it is has no effect.

charypar:
Also I don't think caching is to blame for the fastFT not rendering. It didn't update for some reason even though I saw the fresh JSON content being fetched in the network panel in Chrome dev tools.

PR #192, opened 2015-08-12
Metadata for ad targeting

imranolas:
- [x] UK metadata
- [x] International metadata

`dfp_zone` will be exposed as either `UK` or `US` respectively.


robin-marr-ft:
:+1:

PR #193, opened 2015-08-12
Ads: Update flag and bower dependancies

imranolas:


client/main.js
```diff
@@ -23,7 +23,7 @@ setup.bootstrap(({flags}) => {
 		selector: '.js-video'
 	});
 
-	if(flags.flags.frontPageAds) {
+	if(flags.get('frontPageAds')) {
```
commuterjoy: Are frontpage ads different to any other ads? If not, best us the existing 'ads' flag imho.

matthew-andrews: presumably they're doing this in order to deploy the code without enabling this feature?

imranolas: Yes. Just wanted to QA in relative safety before releasing to live.

charypar:
The build failures are weird, need to investigate in the morning, if it continues...

PR #205, opened 2015-08-17
Separate build-production out into its own step

matthew-andrews:
@charypar 
Making the ‘build’ step separate to the ‘test’ step

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,9 @@ install:
 verify:
 	nbt verify
 
-# FIXME enable verify for ES6 + JSX, test: verify build-production unit-test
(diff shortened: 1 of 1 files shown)


charypar:
@matthew-andrews do we need to do anything about Circle setup or are you handling that also?

charypar:
Build fails because Circle runs `make test && make provision`, so it now doesn't do `build-production`

matthew-andrews:
^^ i think i was just a bit quick to push this branch.  the config in circle should be updated now and this should pass if rebuilt.  i'll hit the button.

PR #207, opened 2015-08-17
synchronise npm & bower versions of express

matthew-andrews:


charypar:
Fails for the same reason as https://circleci.com/gh/Financial-Times/next-front-page/212. It looks like the latest `next-feature-flags-client` is either broken or at least won't build with webpack...

matthew-andrews:
`next-feature-flags-client` is broken. my poor code reviewing skills, sorry!

PR #210, opened 2015-08-19
Fix video scaling on ios6

### LINKED ISSUE #209: Video behaviour is odd on tablet devices (in both safari and chrome on iOS 8)
the small videos get bigger then play (not a major issue apart from the fact that it's unexpected) but the main video gets narrower and smaller then plays (which is not cool, man)

imranolas:
fixes #209


charypar:
@ironsidevsquincy Some of this may be useful for n-video

PR #219, opened 2015-08-22
remove heroku logs command in tidy, now done by nbt

matthew-andrews:
@charypar


charypar:
:+1::heart:

matthew-andrews:
thanks and thanks for the idea @charypar, love this kind of scalpel solution to problems.

PR #226, opened 2015-08-26
Added message prompts client

ifyio:
cc/ @tom-parker @adambraimbridge

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ clean:
 	git clean -fxd
 
 tidy:
-	nbt destroy ${TEST_HOST}
+	#nbt destroy ${TEST_HOST}
 
 provision:
 	nbt provision ${TEST_HOST}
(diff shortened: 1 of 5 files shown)


adambraimbridge:
:+1:

charypar:
@ifyio Can you give a bit more context on what this does? Are those the pop-ups on the bottom we've seen in the Monday meeting or am I getting confused?

Also, if it's a user facing thing, we (home page stream) would appreciate a chance to check with Gadi that he's ok with us shipping it (which I'm sure he is, just want to make sure). Since I am personally unclear on what this is, it makes me uneasy when it just goes in.

ifyio:
@charypar yes sure. It's the javascript / css that sets up and displays the discount message prompt on the site. This message prompt is shown at the bottom left of the page and it is only shown when an anonymous user has been presented with a barrier / product selector and then navigates to another page. It is indeed the popup that we saw in the Monday meeting. Apologies for not giving you guys a chance to look it over. Will make sure to not do that next time around.

charypar:
@ifyio Great thanks!

PR #228, opened 2015-08-28
Use shared color/typography use cases and mixins where applicable

adgad:
/cc @charypar @imranolas @andygnewman @matthew-andrews 

Using common colour use cases and typography mixins where they apply across the site. Hopefully there's nothing too controversial here, but this should massively aide us keep things consistent as thigns change going forward.

No visual differences to homepage.


charypar:
:+1: Awesome!

PR #234, opened 2015-09-01
Fix tag styles on related stories

adgad:
/cc @charypar @imranolas 

Sorry I broke this on Friday!

Before:

screen shot 2015-09-01 at 13 08 47

After: 

screen shot 2015-09-01 at 13 06 55


charypar:
Thanks! ... There's something weird going on with JS smoke tests, getting the same failure on an unrelated branch

PR #243, opened 2015-09-08
Get Editor's Picks from list endpoint

ironsidevsquincy:
cc: @charypar


server/lib/graphql.js
```diff
@@ -14,9 +14,10 @@ const fetch = (backend) => {
 			const now = new Date().getTime();
 
 			console.log('Graphql (', backend.type, ') responded in', now - then, 'ms');
+			if (it.errors) {
+				throw it.errors;
+			}
 			if(it.data) { return it.data; }
-
-			throw it.errors;
```
ironsidevsquincy: previously, this seemed to be swallowing errors...

matthew-andrews:
Is this the first List API thing we're using in Prod??? :smile:

matthew-andrews:
If so, since Lists API is Bronze what's the fallback when it's broken / unupdateable?

charypar:
So after talking to Mark and Gadi, both of them are fine with the section being only on bronze, as long as we can gracefully degrade and hide it when not available. Mark is happy to take care of checking the right content is being displayed when there is content being displayed.

@ironsidevsquincy if you can build in the block hiding into the view level when the section is empty or missing entirely, we can ship this. :ship:

server/graphql/schema.js
```diff
@@ -37,40 +37,13 @@ const queryType = new GraphQLObjectType({
...
-				});
-
-				return directly(3, fetchers)
-				.then(ids => ({
-					title: 'Editor\'s picks',
-					conceptId: null,
-					sectionId: null,
-					items: ids
-				}));
+			type: List,
```
charypar: Can we keep this as a `Collection` and make the interface detect a `List`. I'd like to keep the Collection abstraction, so we can change how we fetch from the backend without necessarily changing the schema, as long as the backend returns a collection we can handle.

server/graphql/schema.js
```diff
@@ -10,7 +10,7 @@ import {
 } from 'graphql';
 
 import {Region} from './types/basic';
-import {Collection, VideoCollection} from './types/collections';
+import {Collection, VideoCollection, List} from './types/collections';
```
charypar: I think `VideoCollection` is an unnecessary import here (and so is list if we make the change below).

PR #246, opened 2015-09-09
Fix mobile tabs not appearing (fix #245)

### LINKED ISSUE #245: Fast FT toggle disappeared on mobile view

charypar:
Caused by linter insisting on rule order that's breaking the CSS override order. Ahh, CSS.


ironsidevsquincy:
:+1:

PR #248, opened 2015-09-09
Highlight dom path based on hash

ironsidevsquincy:
e.g. `/uk#domPath:lead-today | story | title | main-link` displays 

screen shot 2015-09-09 at 15 40 24

Will be useful for tying the CTA dashboard to actual components


server/graphql/backend.js
```diff
@@ -13,7 +13,7 @@ import articleGenres from 'ft-next-article-genre';
 
 // internal content filtering logic shared for ContentV1 and ContentV2
 const filterContent = ({from, limit, genres, type}, resolveType) => {
-	return (items) => {
+	return (items = []) => {
```
ironsidevsquincy: snuck this in, was erroring intermittently on my crummy connection

PR #249, opened 2015-09-10
homepage fix

ifyio:
cc/ @matthew-andrews @charypar @imranolas 

This is a quick fix but I don't have enough context yet to know why `summary` is undefined

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,5 +50,5 @@ deploy:
 clean-deploy: clean install build-production deploy
 
 smoke:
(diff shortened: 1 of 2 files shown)


matthew-andrews:
Piece of causing this is:- 8e46cc50-574d-11e5-a28b-50226830d644

Basically it's because `content.item.summary` is not defined on that piece of content

matthew-andrews:
Update it's because of the slideshow

commuterjoy:
We could look at models in the api client again?

PR #251, opened 2015-09-15
Add scroll depth tracking

ironsidevsquincy:


client/utils/throttle.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+module.exports = (fn, threshold) => {
+	var lastFired;
+	var timer;
+
+	return () => {
+		var now = +new Date();
+		if (!lastFired || lastFired + threshold < now) {
```
charypar: I think the readability of this could benefit from early termination. You could also merge the two cases together as long as you're ok to run the callback on the next tick of the event loop.

```javascript
() => {
    if(timer) return;
    var delay = Math.max(0, threshold + lastFired - new Date());
    
    timer = setTimeout(() => {
        timer = null;
        fn.apply(this, arguments);
        lastFired = +new Date();
    }, delay);
}
```

charypar: Also, if `requestAnimationFrame` is available, we should use it.

client/components/scroll-depth/main.js
```diff
@@ -0,0 +1,29 @@
...
+var throttle = require('../../utils/throttle');
+
+var toArray = nodeList => Array.prototype.slice.call(nodeList);
+
+var track = componentId => {
+	fireTracking('oTracking.event', { category: 'page', action: 'scrolldepth', component: componentId });
+};
+
+var components;
+var scrollHandler = () => {
```
charypar: you could make this a higher order function and have `components` in the closure instead of a module variable (make it slightly more clear where it comes from). But that's minor.

client/components/scroll-depth/main.js
```diff
@@ -0,0 +1,29 @@
...
+	components.forEach((componentEl, index) => {
+		if (componentEl.getBoundingClientRect().top < window.innerHeight) {
+			track(componentEl.dataset.trackable);
+			components.splice(index, 1);
```
charypar: `splice` in a `forEach` doesn't seem safe :) But I might be wrong. Either way, could be rewritten as a `reduce` which `concat`s components that didn't scroll into view yet and skips the others.

client/utils/throttle.js
ironsidevsquincy: Nice. I like it firing immediately, but it does need a tidy

client/components/scroll-depth/main.js
```diff
@@ -0,0 +1,29 @@
...
+
+var toArray = nodeList => Array.prototype.slice.call(nodeList);
+
+var track = componentPos => fireTracking('oTracking.event', { category: 'page', action: 'scrolldepth', componentPos: componentPos });
+
+var scrollHandlerFactory = components => {
+	var componentsViewed = [];
+	return () => {
+		components.forEach((component, index) => {
+			if (component.getBoundingClientRect().top < window.innerHeight && componentsViewed.indexOf(component) === -1) {
```
charypar: It probably makes no difference, but the `.indexOf` is potentially expensive with a lot of components (the algorithm has a complexity of `O(n^2)` now), considering we run this every 250ms when the user is scrolling, could be better done in `O(n)` using something like

```javascript
components = components.reduce((comps, comp) => {
  if(component.getBoundingClientRect().top < window.innerHeight) {
    track(index + 1); // track and don't keep for further checks
  else
    comps.push(comp); // keep for further checks
}, []);
```

ironsidevsquincy: ohh, I like it

charypar: Oh, missed that. You could pass along `[keep, index]` and increment it every step.

charypar:
```javascript
components = components.reduce(([keep, index], c) => {
  if(c.getBoundingClientRect().top < window.innerHeight) {
    track(index + 1);
  else
    keep.push(c);
  return [keep, index + 1];
}, [[], 0]);
```

imranolas: Actually it's available already. The third parameter passed to the reducing function is the `index`. :)

```js
components.reduce((keep, c, index) => {
   // Do stuff
} , [])
```

charypar: Damn! Good spot!

ironsidevsquincy: unfortunately, as the components array changes, the index will change

charypar: Yep, got what you mean, the component index needs to be stable for tracking purposes

PR #253, opened 2015-09-16
Stale cache handling

commuterjoy:
Instructs the cache to update the content every 2 minutes but if the origin is down then show stale content for 3 days.

https://docs.fastly.com/guides/performance-tuning/serving-stale-content

### DIFF
--- a/server/routes/front-page.js
+++ b/server/routes/front-page.js
@@ -11,7 +11,7 @@ export default (region) => {
 
 		res.set({
 			// needs to be private so we can vary for signed in state, ab tests, etc
-			'Surrogate-Control': 'max-age=60'
+			'Surrogate-Control': 'max-age=60,stale-while-revalidate,stale-if-error=259200'
 		});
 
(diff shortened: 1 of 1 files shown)


matthew-andrews:
`stale-while-revalidate` can give some really odd results when different nodes have different versions of content… refreshing the page can swap between nodes…  guardian have 6s `stale-while-revalidate` for their home page… 

```
$ curl -sI http://www.theguardian.com/uk | grep ^Cache
Cache-Control: max-age=57, stale-while-revalidate=6, stale-if-error=864000, private
```

and just 1s for articles:-

```
$ curl -sI http://www.theguardian.com/politics/2015/sep/16/jeremy-corbyn-rules-out-fighting-for-britain-to-leave-eu | grep ^Cache
Cache-Control: max-age=10, stale-while-revalidate=1, stale-if-error=864000, private
```

I suggest we do likewise…

commuterjoy:
I've made it 6s everywhere until I understand why it needs to differ. /cc @phamann ?

charypar:
Does it need to be added to fastft.json route as well? ... Or i could just fix the bug that makes it render on the client even when it has no stories.

PR #254, opened 2015-09-17
Fix the one with fastFT fading out

charypar:
Never render when there is nothing to render. No brainer.

### DIFF
--- a/components/fastft/main.js
+++ b/components/fastft/main.js
@@ -25,7 +25,8 @@ const init = (el, opts) => {
 			items = fastft.items;
 
 			if(document.visibilityState !== 'hidden') {
-				renderFeed(el, items);
+				if(items && items.length > 1)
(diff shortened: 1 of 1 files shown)


components/fastft/main.js
```diff
@@ -35,7 +36,8 @@ const init = (el, opts) => {
 
 	document.addEventListener('visibilitychange', function() {
 		if(document.visibilityState === 'visible') {
-			renderFeed(el, items);
```
ironsidevsquincy: out of interest, why is it matching `visible` here but not matching `hidden` above?

ironsidevsquincy:
:+1:

components/fastft/main.js
charypar: I think it's for browsers that don't have `visibilityState` - `undefined !== 'hidden'`

PR #255, opened 2015-09-17
Dont hang if fastFt endpoint errors

ironsidevsquincy:


charypar:
Oh of course... _facepalm_

:+1:

PR #257, opened 2015-09-18
[Ready] Surface popular topics (flagged)

charypar:
This is for interested people to see what the popular topics API is returning and give hints at what it should be returning. The final design treatment will probably be different

# to do
- [x] fetch topics from Popular API
- [x] allow limiting the amount of topics
- [x] surface in the page
- [x] flag

### DIFF
--- a/server/graphql/backend-adapters/popular-api.js
+++ b/server/graphql/backend-adapters/popular-api.js
@@ -6,7 +6,7 @@ class PopularAPI {
 	topics(ttl = 50) {
 		const url = 'https://ft-next-popular-api.herokuapp.com/topics?apiKey=' + process.env.POPULAR_API_KEY;
 
-		return this.cache.cached(`popular-api.topics`, ttl, () => {
+		return this.cache.cached('popular-api.topics', ttl, () => {
 			return fetch(url)
 			.then((response) => response.json());
 		});
--- a/server/graphql/backend-adapters/popular-api.js
+++ b/server/graphql/backend-adapters/popular-api.js
@@ -1,0 +1,16 @@
+class PopularAPI {
+	constructor(cache) {
+		this.cache = cache;
+	}
+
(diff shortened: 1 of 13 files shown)


server/graphql/backend-adapters/popular-api.js
```diff
@@ -0,0 +1,16 @@
+class PopularAPI {
+	constructor(cache) {
+		this.cache = cache;
+	}
+
+	topics(ttl = 50) {
+		const url = 'https://ft-next-popular-api.herokuapp.com/topics?apiKey=' + process.env.POPULAR_API_KEY;
+
+		return this.cache.cached(`popular-api.topics`, ttl, () => {
```
ironsidevsquincy: no need for backticks

views/partials/lead-today.html
```diff
@@ -1,34 +1,46 @@
 {{#if items}}
 	<section class="flexipod flexipod--lead-today" id="site-content" data-trackable="lead-today" role="region" tabindex="0" aria-labelledby="lead-pod-title">
-		<div class="article-group" role="presentation">
-			{{#slice leads limit=1}}
-				<div class="article no-ff-padding">
-					{{> story-card related=true showSummary=true }}
-				</div>
+		{{#if flags.popularTopics}}
+		<div class="pod-heading" role="presentation">
```
ironsidevsquincy: can put a `data-trackable="popular-topics"` or something on he, so we can group clicks links within it

ironsidevsquincy:
:+1: good stuff, screenshot?

charypar:
topics

charypar:
Needs better design treatment for public release, but it'll do for now until we're happy with what it gives us.

views/partials/lead-today.html
charypar: It's on the `<ul>` below

PR #258, opened 2015-09-21
Bump to node 4.1.0

ironsidevsquincy:


charypar:
:+1:

imranolas:
You might also want to update the babel transforms, so that we're only transforming features that aren't natively supported.

charypar:
@imranolas good point, any good docs for that?

ironsidevsquincy:
oh, yeah, that's cool. I'll update

imranolas:
Here's a list of the natively supported es6 features. https://nodejs.org/en/docs/es6/
These are the babel transformers. https://babeljs.io/docs/advanced/transformers/

You should be able to use the `whitelist` option to declare *only* the transformations you want to apply. It starts to get a little manual at this point because babel doesn't yet have the option of choosing a compile target other than es5.

circle.yml
```diff
@@ -2,7 +2,7 @@ machine:
   services:
     - docker
```
matthew-andrews: can also delete `docker`

server/app.js
```diff
@@ -1,2 +1,13 @@
-require('babel/register');
+require('babel/register')({
+	whitelist: [
+		'es6.arrowFunctions',
```
ironsidevsquincy: annoyingly need this, or else this wonderful error is produced

screen shot 2015-09-21 at 14 30 32

PR #266, opened 2015-09-23
Use semver ^ to avoid pulling in breaking changes

keirog:
Preparing for an o-grid upgrade with some breaking changes in it. I imagine front-page will be ok, but better to be safe.


charypar:
:+1:

PR #271, opened 2015-09-28
Use NBT v4

ironsidevsquincy:
Still need our own `.eslintrc`, as using modules and jsx


charypar:
:+1:

matthew-andrews:
@ironsidevsquincy do you think we should put the `modules` flag into the NBT build tools `eslintrc` file?

ironsidevsquincy:
@matthew-andrews problem is, it screws with the `use strict` directive; with `modules` on, eslint doesn't like you using `use strict`, which is fair as modules don't need it, but we're not using modules everywhere, (i.e. node can't yet), so they're still commonJS modules which need `use strict`. 

this is a real issue if we're using modules on the client (which we are as it's babel), and commonJS on the server, but using the same eslint config for both

matthew-andrews:
ah ok

Makefile
```diff
@@ -3,21 +3,23 @@ TEST_URL := "http://ft-next-fp-branch-${CIRCLE_BUILD_NUM}.herokuapp.com"
...
+smoke:
+	nbt test-urls ${TEST_HOST}
+	export TEST_URL=${TEST_URL}; nbt nightwatch test/browser/tests/*
 
-run-local:
-	nbt run --local
+test: verify unit-test
 
 build:
 	webpack
```
matthew-andrews: `/[ON]BT4/` is built on webpack rather than browserify so you should be able to remove the webpack integration code now, if you want :smile:  @ironsidevsquincy @charypar

ironsidevsquincy: good point

PR #273, opened 2015-09-28
Webpack hot loading

charypar:
This changes the way we use webpack to enable hot module replacement (HMR) for both JS and CSS in preparation of the Layout engine prototype, to let us iterate quickly on it.  

To enable this, `make watch` now runs a separate webpack server (a node server running two webpack middlewares - dev and hot), which serves the JS and CSS bundle (actually just the JS bundle, as CSS is included as style tags in development to enable the hot module replacement) and also an endpoint with notifications of the bundle being rebuilt on changes.

The main application proxies to the webpack server on the `/front-page/` path, to keep the asset paths consistent with `next-express`, the only difference is the CSS bundle endpoint in the webpack server returns a 404. The bundle also includes a client for the HMR endpoint, which listens to the server sent events on bundle change and hot-swaps the changed JS modules with new ones.

To leverage the HMR for CSS too as much as possible, the React component styles are now required from their respective JS files, isolating them from the main huge style bundle and expressing the dependencies how they really are (i.e. component JS code depends on its CSS styles).

The config includes special treatment of React component modules to enable the react component hot loading. This lets you swap out any component code and replace it with a changed one, live, without page refresh, while keeping the state (`props` and `state`) of the component intact (e.g. list of fastFT articles, or the resulting state of a couple steps of user interaction). The end result is a feedback loop of about 5 seconds (there's a lot of javascript and CSS) between saving a file and seeing the rendered result of the code on screen. Neat.

## To Do

- [x] hot loading server
- [x] react transform HMR + errors
- [x] integrate into the Makefile
- [x] make sure `make build` and `make build-production` still works correctly


client/main.js
```diff
@@ -1,4 +1,4 @@
-import fastFT from '../components/fastft/main';
+/* global document */
```
ironsidevsquincy: is this needed? - https://github.com/Financial-Times/next-build-tools/blob/master/config/eslint.json#L3

ironsidevsquincy:
:+1:

client/main.js
charypar: Isn't. Removing.

PR #283, opened 2015-10-07
Make layout more abstract and content & configuration driven

charypar:
This will move all the configuration of the entire page into a single configuration object pasded top down, which can be produced by a layout engine and manipulated from outside, as a step towards

a) interactive prototype to demonstrate layout capabilities
b) intelligent story volume driven layout engine

## To do

- [x] Drive page sections with configuration too
- [x] Single configurable `Section` component
- [x] Single configurable `SectionContent` component
- [x] Unified `Card` component with `props` driven components and style
- [x] improve styles to implicitly handle tall cards


components/fast-ft/fast-ft.js
```diff
@@ -3,13 +3,15 @@ import FastFtItem from './fast-ft-item';
...
 	render() {
-		const articleEls = this.props.articles.map(article => <li key={article.id}><FastFtItem article={article} /></li>);
+		const articleEls = this.props.articles.slice(0, 6).map(article => <li key={article.id}><FastFtItem article={article} /></li>);
 		return (
-			<div className="fast-ft">
-				<h2><span className="fast-ft__logo fast-ft__logo--fast">fast</span><span className="fast-ft__logo fast-ft__logo--ft">FT</span></h2>
-				<ol className="fast-ft__items">
-					{articleEls}
-				</ol>
+			<div style={{display: 'flex'}}>
```
ironsidevsquincy: `className="fast-ft-wrapper"` and bung in the css

ironsidevsquincy:
:+1:

components/fast-ft/fast-ft.js
charypar: I know, the sidebar treatment needs more thinking, it's not gonna stay in like this, because it's logically part of the `Section` rather than the sidebar - it's a card container similar to the `SectionContent`, but we don't have it as an explicit concept, so I just smashed this in as a temp fix.

Will extract to CSS with some big HACK comment :)

PR #284, opened 2015-10-07
Use graphql api service

ironsidevsquincy:


package.json
```diff
@@ -8,14 +8,10 @@
     "cheerio": "^0.19.0",
     "dateformat": "^1.0.11",
     "directly": "^1.0.2",
-    "es6-promise": "^2.1.1",
     "forever": "^0.14.2",
-    "ft-api-client": "^4.0.1",
     "ft-next-article-genre": "^1.4.0",
     "ft-next-article-primary-tag": "1.0.1",
```
charypar: `ft-next-article-*` can go as well, they are only used in the GraphQL layer

package.json
```diff
@@ -8,14 +8,10 @@
     "cheerio": "^0.19.0",
     "dateformat": "^1.0.11",
     "directly": "^1.0.2",
```
charypar: directly can likely go as well.

server/libs/get-data.js
```diff
@@ -0,0 +1,27 @@
+import { logger } from 'ft-next-express';
+
+const getData = query => {
+	return fetch(`http://next-graphql-api.ft.com/?apiKey=${process.env.GRAPHQL_API_KEY}`,
+		{
+			method: 'POST',
+			body: JSON.stringify({
+				query: query
```
charypar: We need to send up flag overrides. Either all of them, or at least the CAPI one.

@matthew-andrews is this potentially something worth doing on `fetch` level...? Seems like a recurring problem with service extraction.

server/routes/fastft.js
```diff
@@ -1,19 +1,13 @@
-import graphql from '../lib/graphql';
 import queries from '../config/queries';
```
charypar: `import {fastFTQuery} from '../config/queries'` and use it below?

server/routes/front-page.js
```diff
@@ -1,17 +1,11 @@
-import graphql from '../lib/graphql';
 import queries from '../config/queries';
```
charypar: And similar to `fastFTQuery` above here...

server/routes/query.js
```diff
@@ -1,17 +1,16 @@
-import graphql from '../lib/graphql';
```
charypar: We can get rid of this route entirely, IMO

charypar:
:+1:

PR #286, opened 2015-10-08
[Ready] Layout prototype UI

charypar:
This will add UI to exercise the various tools the grid layout engine supports live in the browser to demonstrate the flexibility of the layout system and make it easier for people to understand the various degrees of freedom we have in the design. 

Eventually, when the simpler weight based layout engine is built, this will get updated to control the story weights instead (and also randomise them to a degree to get different layout templates).

# To Do

- [x] client-side rendering
- [x] make sure hot loading works
- [x] layout controls overlay
- [x] card options controls
- [x] enforce column and width constraints when changing values
- [x] fetch related stories for everything
- [x] change all taxon colours to claret

### DIFF
--- a/components/layout-overlay/layout-overlay.js
+++ b/components/layout-overlay/layout-overlay.js
@@ -1,5 +1,7 @@
 import React, {Component} from 'react';
 
+import Section from './section/section';
+
 export default class LayoutOverlay extends Component {
 	constructor(props) {
 		super(props);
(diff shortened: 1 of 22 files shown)


components/layout-overlay/layout-overlay.js
```diff
@@ -0,0 +1,51 @@
...
+
+		this.state = {expanded: false};
+	}
+
+	toggle(e) {
+		e.preventDefault();
+		this.setState({expanded: !this.state.expanded});
+	}
+
+	updateCards(idx) {
```
ironsidevsquincy: `idx`? a naming convention?

ironsidevsquincy:
:+1:

components/layout-overlay/layout-overlay.js
charypar: Short for `index`, should I rename...?

PR #289, opened 2015-10-08
Fix live blog cards

ironsidevsquincy:
### Before

screen shot 2015-10-08 at 17 03 48

### After

screen shot 2015-10-08 at 17 03 42


charypar:
:+1:

PR #292, opened 2015-10-13
Layout prototype constraints

charypar:
This limits the options in the layout UI slightly to only sane combinations based on @clmntnbrn's input. Also adding a bit of styling to make it look prettier.

- [x] New title size for the lead card without an image
- [x] Tag, title and stand first size governed by a single font size
- [x] Width 5 and 6 only allowed in the first column
- [x] card borders to aid visually distinguishing sections

### DIFF
--- a/components/card/card.js
+++ b/components/card/card.js
@@ -8,14 +8,10 @@ import Related from './related/related';
 
 import Ad from './ad';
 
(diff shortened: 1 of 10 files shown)


ironsidevsquincy:
:+1:

PR #293, opened 2015-10-14
Header markets data

ironsidevsquincy:
Behind flag, markets not fully finalised yet

screen shot 2015-10-14 at 12 29 26

### DIFF
--- a/server/routes/front-page.js
+++ b/server/routes/front-page.js
@@ -18,17 +18,30 @@ export default (region) => {
 
 		getData(query, res.locals.flags)
 			.then(data => {
-				res.render('front-page',
-					{
-						layout: 'wrapper',
-						FastFtFeed,
-						Feed,
-						Layout,
-						content: data,
-						initialLayout,
-						region
-					}
-				);
+				const renderParams = {
+					layout: 'wrapper',
+					FastFtFeed,
+					Feed,
+					Layout,
+					content: data,
+					initialLayout,
+					region
+				};
+				if (res.locals.flags.frontPageHeaderMarketsData) {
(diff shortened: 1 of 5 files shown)


server/routes/front-page.js
```diff
@@ -18,17 +18,30 @@ export default (region) => {
...
+					layout: 'wrapper',
+					FastFtFeed,
+					Feed,
+					Layout,
+					content: data,
+					initialLayout,
+					region
+				};
+				if (res.locals.flags.frontPageHeaderMarketsData) {
+					renderParams.header = {
```
charypar: can you extract that before object literal to a variable above for a bit better readability?

ironsidevsquincy:
Future task: move this into the server - whether it's content/enhanced is debatable, but the latency in it appearing at the top of the page mean users miss it, and the extra (async) server request should add a minimal performance hit

charypar:
:+1: as long as we revisit it. 

The latency is minimal on good network on desktop, real-world mobile cases will be orders of magnitude worse, whereas a concurrent server-side request will cost us next to nothing, because the markets proxy is faster than all the content APIs we need to talk to in order to build the page. And if we pull in the markets data through GraphQL, we can even use the same caching/polling we do there, fetch in in one request with the other content and not affect the page load times and end user performance at all.

Another point I missed is that client-side JS isn't for free either, running a bunch of JavaScript to add content to the page, especially on weaker mobile devices, will affect the perceived performance even worse than the high-latency network.

ironsidevsquincy:
https://github.com/Financial-Times/next-front-page/issues/294

PR #296, opened 2015-10-15
Add markets data disclaimer

ironsidevsquincy:
screen shot 2015-10-15 at 11 48 59


charypar:
Html strings. Html strings everywhere...

:+1: ship it.

PR #299, opened 2015-10-15
FastFT item & time display order

andygout:
Changes to prototype FastFT

- [x] Styling: order of content changed to item text then time
- [x] Styling: 1px left border (like Opinion cards)
- [x] Check timestamps are local
- [x] Amend time to 24 or 12 hour clock, as required (changed to 24 hour clock)

@clmntnbrn

### DIFF
--- a/components/fast-ft/fast-ft-item.js
+++ b/components/fast-ft/fast-ft-item.js
@@ -4,7 +4,7 @@ import {format as dateFormat} from 'o-date';
 class FastFtItem extends Component {
 	render() {
 		const article = this.props.article;
-		const time = dateFormat(article.lastPublished, 'H:mm a');
+		const time = dateFormat(article.lastPublished, 'H:mm');
 		return (
 			<div className="fast-ft__item">
 				<h2 className="fast-ft__item__title">
--- a/components/fast-ft/fast-ft-item.js
+++ b/components/fast-ft/fast-ft-item.js
@@ -7,9 +7,10 @@ class FastFtItem extends Component {
 		const time = dateFormat(article.lastPublished, 'H:mm a');
 		return (
 			<div className="fast-ft__item">
-				<time className="fast-ft__time" datetime={article.lastPublished}>{time}</time>
 				<h2 className="fast-ft__item__title">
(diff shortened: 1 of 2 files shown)


andygout:
fastft_item_time_order_display screen shot 2015-10-15 at 17 16 59

clmntnbrn:
@andygout good to go!

components/fast-ft/fast-ft-item.js
```diff
@@ -7,9 +7,10 @@ class FastFtItem extends Component {
 		const time = dateFormat(article.lastPublished, 'H:mm a');
 		return (
 			<div className="fast-ft__item">
-				<time className="fast-ft__time" datetime={article.lastPublished}>{time}</time>
 				<h2 className="fast-ft__item__title">
 					<a className="fast-ft__item__title-link" href={'/content/' + article.id} data-trackable="link">{article.title}</a>
+					<span> – </span>
```
ironsidevsquincy: why wrapped in a `span`?

charypar: Because whitespace... It's a slightly annoying react thing, it needs nodes to put ids on them for reconciliation, so it would wrap it in a span anyway (in case you have a `element text element` situation), just with wrong spaces around.

andygout:
Just changed my computer's time zone to Los Angeles time (8 hours behind) and times updated appropriately.  Quite how we test I am not sure but will have a think.

screen shot 2015-10-15 at 17 31 29

charypar:
@andygout Not sure how to automatically test that either, but wouldn't worry about it for a prototype.

components/fast-ft/main.scss
```diff
@@ -6,6 +6,7 @@
 .fast-ft {
 	background-color: getColor('pink');
 	padding: 10px;
+	border-left: 1px solid $next-pink-dark;
```
charypar: instead of the `$next-pink-dark` can you use `getColor`?

charypar: `getColor('warm-3')` like other cards

charypar:
:+1:

PR #308, opened 2015-10-16
Throw a 500 if there is no content to show

charypar:
... rather than showing an empty page.

@ironsidevsquincy @matthew-andrews is it worth also keeping a copy of the last fetched content and serving that for a certain time if we get an empty or no response from the backend, or is it ok to rely on Fastly to serve stale?

### DIFF
--- a/server/routes/front-page.js
+++ b/server/routes/front-page.js
@@ -7,6 +7,11 @@ import Feed from '../../components/feed/feed';
 import Layout from '../../components/layout/layout';
 import initialLayout from '../../components/layout/config';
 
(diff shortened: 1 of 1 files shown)


matthew-andrews:
:-) :+1:

matthew-andrews:
I'm fine with relying on fastly … seems simpler

matthew-andrews:
but if you think it's worth it then can keep the last known good

matthew-andrews:
(i dont mind)

charypar:
Ok, let's go with just 500 and if it becomes a problem, solve then.

Issue #265, opened 2015-09-23
Subheads should not be marked up as H2


i-like-robots:
First reason is that it's [not the recommended way to markup subheadings](http://www.w3.org/html/wg/drafts/html/CR/common-idioms.html#sub-head) but more importantly it makes navigating the page for a screen reader user more difficult (if they're skipping header to header).

Usage is here:
- https://github.com/Financial-Times/next-front-page/blob/master/views/partials/story-card.html#L27 
- https://github.com/Financial-Times/next-front-page/blob/master/views/partials/story-card.html#L31
- https://github.com/Financial-Times/next-front-page/blob/master/views/partials/carousel-card.html#L13

PR #309, opened 2015-10-16
Amend: Subheads markup from h2 to p

andygout:
Fixes https://github.com/Financial-Times/next-front-page/issues/265

Link in issue was broken so referenced this: http://www.w3.org/TR/html5/common-idioms.html#sub-head


ironsidevsquincy:
:+1:

PR #311, opened 2015-10-19
Switch current home page to n-image

### LINKED ISSUE #131: Responsive image loading
As a performance enhancement, we should probably decide on an image size to fetch based on the device screen size - no point fetching a 720 pixel image if it's only ever gonna get used at the width of 240.

This should get us a better full page load speed on mobile devices.

charypar:
To improve page load time on small screens and also hopefully get better ordering of asset loading (icons and fonts first, images second).

Fix #131

# To do

- [x] switch to n-image
- [x] define breakpoint sizes
- [x] make the sizes fit the image cases (lead, top stories story, editor's pick, others)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,7 @@
     "n-video": "^2.8.2",
     "n-message-prompts": "Financial-Times/n-message-prompts#^v1.0.0",
     "n-layout": "^2.0.0",
-    "n-image": "~1.3.1"
+    "n-image": "^1.3.1"
   },
   "resolutions": {
     "o-ft-icons": "next",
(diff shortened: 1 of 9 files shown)


bower.json
```diff
@@ -7,7 +7,8 @@
     "o-date": "^2.0.0",
     "n-video": "^2.8.2",
     "n-message-prompts": "Financial-Times/n-message-prompts#^v1.0.0",
-    "n-layout": "^2.0.0"
+    "n-layout": "^2.0.0",
+    "n-image": "~1.3.1"
```
adgad: Should be ^

charypar: Oh, obvs. Good catch

adgad:
:+1:

PR #314, opened 2015-10-20
Fix: ESLint problems

andygout:
@ironsidevsquincy


charypar:
:+1:

charypar:
Make sure the production build runs properly and minifies the JS and CSS (i.e. that the plugins get added)

PR #316, opened 2015-10-20
Add timeout to graphQL fetch

adgad:
Should prevent front page response times ballooning if graphQL has problems


charypar:
:+1:

PR #318, opened 2015-10-20
Adding health checks

ironsidevsquincy:
screen shot 2015-10-20 at 15 31 37


server/config/health-checks/pingdom.js
```diff
@@ -0,0 +1,36 @@
+export default {
```
matthew-andrews: sorry about this…  i think the pingdom checks shouldn't be added to `__health` because they can directly feed into the platinum alert route into PagerDuty.

server/config/health-checks/graphql.js
```diff
@@ -0,0 +1,151 @@
...
+                            ... Basic
+                            ... Extended
+                        }
+                    }
+                }
+		    `
+        },
+        {
+            name: 'FastFT',
+            severity: 1,
```
matthew-andrews: Should be 2 or lower — CAPI v2 is currently bronze

server/config/health-checks/pingdom.js
ironsidevsquincy: I did check with @aintgoin2goa, he recommended added pingdom to here (and pretty much everything else health-related), as it's a single place to get an overview of everything, which make sense to me. But if that's not the general policy, I'm easy either way

matthew-andrews: ok let's us three chat tomorrow and make a plan for everything…  my concerns are:-

- pingdom can talk to pagerduty directly -> more informative alerts
- if pingdom is in `__health` then both will fire if the pingdom alerts fail -> double alerts for the same thing… potentially confusing

server/config/health-checks/graphql.js
```diff
@@ -0,0 +1,151 @@
...
+                            id
+                            url
+                            taxonomy
+                            name
+                        }
+                    }
+                }
+
+                query TopStoriesTest {
+                    popularTopics {
```
matthew-andrews: what's to stop these drifting out of date with the actual queries that generate the front page?

ironsidevsquincy: No really concerned about the actual data it gets, just that it can make the query (the data is for the top three sections)

If those sections change, then the tests will also need to

charypar:
Should top stories stay at severity 1 (data is coming through CAPIv1)?

charypar:
:+1:

server/config/health-checks/pingdom.js
charypar: I think I agree, we should get rid of these.

PR #321, opened 2015-10-20
Try to go without `bower` as a dev dependency

matthew-andrews:


matthew-andrews:
I don't think this is failing because of bower.

charypar:
@matthew-andrews What's this for? Is it still relevant?

matthew-andrews:
Yes, apps shouldnt really have `bower` as a dev dep.  Let me try rebasing.

ironsidevsquincy:
This was just while obt was playing up

charypar:
in that case :+1:

matthew-andrews:
success!!!

matthew-andrews:
success!!!

PR #324, opened 2015-10-22
Hide alt text on images

charypar:
All of them are illustrational.

@i-like-robots

### DIFF
--- a/server/config/queries.js
+++ b/server/config/queries.js
@@ -18,7 +18,6 @@ const fragments = `
 		}
 		primaryImage {
 			rawSrc
-			alt
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


ironsidevsquincy:
:+1:

i-like-robots:
On the image elements themselves (if you can) it would be good to add `role="presentational"` just so any screen reader user isn't subjected to their software trying to interpret the filename into something.

charypar:
@i-like-robots would have to do it in n-image. But what it does is leave `alt` empty, which I think is equivalent?

i-like-robots:
In theory but not in reality - yet =(

charypar:
Issue on n-image then?

PR #328, opened 2015-10-23
[Ready] Make layout engine responsive

charypar:
Adds a layout compiler which takes a layout description per o-grid layout and transforms it into a set of cards fulfilling the layouts with a minimal set (well almost minimal set) of markup and classes / o-grid colspans.

Cards get rendered once if they are always in the same column, multiple times when they are in different columns in different layouts. The other properties of cards also have different values per layout and are implemented with mobile-first CSS classes targeting the layouts (e.g. `tag--M--large` applies to `M`, `L` and `XL` layouts and `tag--large` is a default).

# to do
- [x] responsive card sizes
- [x] aggregate colspans & card level responsive hide/show
- [x] responsive card options
- [x] correct rendering of columns
- [x] correct rendering of card options
- [x] layout switch in the editing UI
- [x] update UI options
- [x] responsively align the image to the bottom of the card when there are no stories
- [x] don't distribute whitespace evenly in columns
- [x] adding / removing cards
- ~~advert card~~ needs rethinking, will do as a separate PR


ironsidevsquincy:
Why not :+1:

PR #329, opened 2015-10-26
Increase timeout to 4000

adgad:
/cc @charypar @matthew-andrews @ironsidevsquincy 
The max response time for graphql sometimes pops between 3000-3500ms. This should only really be a temporary placation, until we do the polling of graphQL.


charypar:
:+1:

PR #330, opened 2015-10-26
Poll graphql for data

adgad:
/cc @matthew-andrews @charypar @ironsidevsquincy @andygout 
Polls graphql every minute for front page data, instead of on request.

* ~~`app.listen` and healthcheck polling waits for initial fetch to complete before starting. If we fail to get content on startup, then the app will not start up (and build won't pass etc).~~ Decided not to do this: if heroku restarts but fails to get any content, the new app won't ever start listening. 
* Stale polled content will remain in memory indefinitely (until the app next restarts) - with healthchecks that escalate as the content gets 10/30 minutes out of date. 
* Added a US top stories healthcheck
* Set caching for FastFT Json to same as front page HTML (to cache on Fastly and stale on error)

TODO
- [x] Get mockFrontPage working
- [x] Check error handling
- [x] Add healthcheck that checks the freshness of content


server/init.js
```diff
@@ -70,6 +71,9 @@ app.get('/home/fastft.json', fastft);
 const port = process.env.PORT || 3001;
 
 export default app;
-export let listen = app.listen(port, () => {
-	logger.info('Listening on ' + port);
-});
+export let listen = startPolling().then(() => {
+	return app.listen(port, () => {
```
adgad: There's a small chance this may lead to flaky builds (or at least builds failing when content sources are down) - but I suspect that's a good thing, so we keep an app with data already in it.

server/routes/fastft.js
```diff
@@ -1,18 +1,21 @@
-import { fastFT as fastFTQuery } from '../config/queries';
-import getData from '../libs/get-data';
+import { getData } from '../libs/graphql-poller';
 import { logger } from 'ft-next-express';
 
-module.exports = function (req, res, next) {
-	res.set({
-		'Cache-Control': 'max-age=20, public' // 20 seconds
```
adgad: @charypar Was there any reason why this was such a small (and public) cache? Can you forsee any problems changing it to the same as homepage (cache on fastly only)

charypar: Originally the poll frequency on the front-page was 30 seconds, but it put too much load on the router, I believe. 

I actually think we could get rid of this route and go straight from the client to `graphql-api` if we implement CORS or JSONP.

charypar:
:+1:

PR #331, opened 2015-10-27
Slightly safer to do hashed assets before shipping

matthew-andrews:
We're protected by preboot anyway but just to be on the safe side


charypar:
:+1:

PR #335, opened 2015-10-29
Fix bad font file reference

ironsidevsquincy:
By using o-fonts v1.8.4

e.g. fixes

screen shot 2015-10-29 at 14 29 37


charypar:
:+1:

Issue #336, opened 2015-10-30
New Homepage layout click tracking should be consistent with the old ones


adgad:
The dom-paths should (ideally) match up between the old and new layouts, so that it's easier to compare when AB testing.

e.g. Old: `lead-today | story | title | main-link`
New: `card | link`

PR #339, opened 2015-11-03
Get (most) tracking for prototype in sync with old page

adgad:

"Fixes" https://github.com/Financial-Times/next-front-page/issues/336 - with some differences.

* I prefer `top-stories` to `lead-today`as the trackable for top stories

* The `title` seems a bit superfluous in the example in the issue, so ditching that. The resulting trackable is: `opinion | card | main-link` on the new prototype.

I think it's fine to handle these differences in beacon as needed for now.


ironsidevsquincy:
:+1: on making the trackables more useful

PR #340, opened 2015-11-04
Bump to n-layout v4 and fix dep resolutions

phamann:
We've introduced a breaking change to the `myft-client` which powers the save buttons on the homepage. This bumps to latest version. Sadly the dep resolutions are all needed but should only be temporary, I will clean up tomorrow.


adgad:
:+1:

phamann:
This needs to be rolled out as part of a phased deployment so won't merge just yet.

PR #350, opened 2015-11-10
Make IE8 look less terrible

adgad:
* `o-grid-container` falls back to a fixed width, so removed uses of this within components that should be less than full width
* This enables fastFT and most popular content to be accessed on the front page, and makes things look less terrible
* Also only apply overflow-x on Small breakpoints since this is needed on pages that don't support media queries (needs to be done in next-sass-setup also) 

ie8


ironsidevsquincy:
:+1:

PR #352, opened 2015-11-11
FastFT/Top Stories tabs

ironsidevsquincy:
Basically copy existing next front page (will need to be revisited re: `target` pseudo-class)

screen shot 2015-11-11 at 16 25 18

(licecap's gone screwy, so just a static image)


adgad:
:+1:

PR #355, opened 2015-11-12
Make fastft not go through preflight

adgad:
/cc @commuterjoy @ironsidevsquincy (Needs service registry merged first)

### DIFF
--- a/server/init.js
+++ b/server/init.js
@@ -68,7 +68,7 @@ app.get('/front-page', frontPage('UK'));
 app.get('/international', frontPage('US'));
 app.get('/uk', frontPage('UK'));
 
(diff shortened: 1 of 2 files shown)


server/init.js
```diff
@@ -68,7 +68,7 @@ app.get('/front-page', frontPage('UK'));
 app.get('/international', frontPage('US'));
 app.get('/uk', frontPage('UK'));
 
-app.get('/home/fastft.json', fastft);
+app.get(/\/(__)?home\/fastft.json/, fastft);
```
ironsidevsquincy: `fastft\.json`

ironsidevsquincy:
:+1:

server/init.js
adgad: good spot!

PR #359, opened 2015-11-13
Use existing tab nav on new layout

ironsidevsquincy:
Same markup and styling (though it's pretty horrible so could do with a refactor)


adgad:
:+1:

PR #367, opened 2015-11-17
Add cols to config

ironsidevsquincy:
To create

screen shot 2015-11-17 at 15 13 50


charypar:
Could make those objects instead of pattern strings to make it easier to edit if necessary and also validate and generally work with. Data all the things! :) #drivebycomment

ironsidevsquincy:
Ha! As always, quite right

PR #376, opened 2015-11-20
Remove: Inline display from card image link class

andygout:
cc @ironsidevsquincy 

Fix to make landscape cards work on IE10


ironsidevsquincy:
:+1:

PR #382, opened 2015-11-20
Fix hover state of fastFT logo

matthew-andrews:
This is what this line was intended to do… oops!


matthew-andrews:
wait no this isn't right

PR #383, opened 2015-11-20
Fix underline of fastFT logo

matthew-andrews:
This is what this line was intended to do… oops! [take 2]


ironsidevsquincy:
ah, yeah. we use borders for out link underlines

:+1:

PR #394, opened 2015-11-24
Try out React branch of NBT

matthew-andrews:
[ Even less code in `.eslintrc` — the only thing that remains is the modules … which we might like to allow in NBT or hold off until Node natively supports it ? ]


matthew-andrews:
Oh actually NBT already has `modules: true`

andygout:
:+1:

Issue #371, opened 2015-11-18
styling tweaks for front page


clmntnbrn:
Styling things we need to do on the front page (this is an evolving list!):

 - [x] tighten up the margin above the top stories section
 - [x] equally distribute any excess height needed between the cards in 1 column (instead of just on the last one)
 - [x] have 4 videos not 6
 - [x] create the rule that:
- IF		there is a story package with the lead story with 3+ related links
- THEN	we show 3 related links
- AND		move the second story to the top of col 1
- AND		move the last story of col 1 to the top of col 2 and change it’s font to ‘Tiny’
- AND 	lose last story on col 2
 - [x] make the image stay under the text when the card grows, rather than being at the bottom
 - [x] this ^ is the opposite for editor’s picks
 - [x] change editors picks title font from Financier to Metric (18/20px)
 - [x] change Financier Italic standfirst (opinion section) to Metric Light
 - [x] use the proper quote icon (opinion section)
 - [x] move author names and quote marks to under the title (opinion section)
 - [x] swap images for headshots on articles with bylines (opinion section)
 - [ ] adjust the spacing of markets data (we can fit more in on smaller breakpoints)
 - [x] implement new section heading style as per latest design on the wall
 - [x] properly align wrapped lines on related links
 - [x] Align the nav to the main content (but leave the bg full width)
 - [x] I'm not sure the format of the tiny cards are working... just sayin'
 - [x] change tag to MetricWeb font-weight: 600;
 - [x] change standfirst font size to 18px at M breakpoint (at the moment it's 22px, the same as the headline)
 - [ ] hover effect styling to buttons (subnav & search toggle / subnav parent topic buttons), i.e. not browser default blue border employed by some browsers (inc. Chrome)

These may not be considered part of this work, but just to note them:
 - [ ] Increase contrast of search background
 - [ ] home selected state is still visible behind the search box when expanded

PR #396, opened 2015-11-24
Ability to stick images

ironsidevsquincy:
And stick headshots by default

screen shot 2015-11-24 at 16 22 36


andygnewman:
Can we use consistent quote symbols for brands - we're using the pullquote from o-ft-icons/next on the branded articles https://github.com/Financial-Times/o-ft-icons/blob/next/svg/pullquote.svg

ironsidevsquincy:
absolutely, on our to do list - https://github.com/Financial-Times/next-front-page/issues/371

PR #404, opened 2015-11-26
IE-friendly landscape cards

andygout:


ironsidevsquincy:
:+1:

components/card/main.scss
```diff
@@ -15,6 +15,7 @@
 	flex-direction: column;
 	line-height: 1;
 	flex: 1;
+	-ms-flex-preferred-size: auto;
```
ironsidevsquincy: not needed anymore

PR #405, opened 2015-11-26
Correct: Section titles (spelling & wording)

andygout:
cc @ironsidevsquincy

### DIFF
--- a/components/layout/config.js
+++ b/components/layout/config.js
@@ -243,7 +243,7 @@ export default [
 	},
 	{
 		id: 'technology',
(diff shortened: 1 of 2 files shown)


components/layout/config.js
```diff
@@ -336,7 +336,7 @@ export default [
 	},
 	{
 		id: 'lifestyle',
```
andygout: Should `id` and `style` be changed to `life&arts`/`lifeandarts`?

ironsidevsquincy:
:+1:

PR #418, opened 2015-11-30
Style tweaks

adgad:
/cc @ironsidevsquincy 
* Header styles (needs verifying font size etc)
* Tag font weight
* M layout font/image tweaks
* Removing duplicate markets config

screen shot 2015-11-30 at 14 38 30


ironsidevsquincy:
:+1:

PR #420, opened 2015-11-30
Change top stories layout if top story has related content

adgad:
/cc @ironsidevsquincy 

Allow for config overrides based on conditions on the items. Not sure if this is clever or hacky/confusing, but can't think of a better way to do it!


ironsidevsquincy:
:+1:

PR #421, opened 2015-11-30
don't minimize css before autoprefixing

wheresrhys:
Because otherwise autoprefixer config comments get stripped out before they are needed.

The css that's output is minified but god knows why. Don't know if css-clean is being applied either. 

Can we move front-page back to a standard build process please? (I agree it's worth us considering moving away from obt though)

@ironsidevsquincy @adgad

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -40,9 +40,9 @@ const config = {
 				test: /\.scss|sass?$/,
 				loader: ExtractTextPlugin.extract(
 					[
-						'css?sourceMap&minimize',
+						'css?sourceMap',
 						'autoprefixer',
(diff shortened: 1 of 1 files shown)


matthew-andrews:
what's wrong with obt? :-/

wheresrhys:
one for the dev huddle

matthew-andrews:
:cry:

wheresrhys:
but it's not the tool, more the work flow and the fact we can't 'move fast and break stuff' with our build tool as it has to support a wider community

ironsidevsquincy:
Actually, I added as a point last week

Want to move back, but will need to split the css out of the js first. Also, need to figure out what's needed for the layout overlay tool

PR #423, opened 2015-12-01
More style fixes

adgad:
/cc @ironsidevsquincy 
* Technology/Markets/Life & Arts cards will be landscape on L upwards
* Only do the related links reorganisation if there are 3 or more related links


ironsidevsquincy:
:+1:

PR #428, opened 2015-12-01
No need to pass initial state into Fast FT

matthew-andrews:
It only needs to re-render when it gets a response back from the `.json` endpoint.


matthew-andrews:
This might do something actually

charypar:
:+1: :)

matthew-andrews:
I'm not so sure about this anymore.  I _suspect_ there's something in the way the animation library works that relies on this ……  Still figuring it out …

charypar:
It should be able to pick up the server-rendered markup and just carry on. What seems to be wrong?

matthew-andrews:
If you re-render with different content than was produced on the server it will throw this warning:-
screen shot 2015-12-02 at 08 33 36

This is a problem because it's reasonably likely as by the time `render` runs for the first time the state will have changed (new article, etc).

The user impact of this is that the whole widget re-renders (and therefore there are no animations)

Thing is, it feels like it *should* be able to cope …

matthew-andrews:
looking on the internet it seems like that the only way to do it is to pass the initial state into React :-/

PR #431, opened 2015-12-01
Make blackberry 7 slightly less terrible

adgad:
/cc @wheresrhys @ironsidevsquincy @TheoLeanse @andygout 

* Apply the autoprefix fix on all front-page uses of flexbox
* Add a clearfix to the content container, to prevent footer overflowing (will probably move this to n-layout (around the footer itself) shortly)


ironsidevsquincy:
:+1:

PR #435, opened 2015-12-02
Use standard build process

ironsidevsquincy:
Rather than using our own webpack build

Means we lose the ability to include css via js, but the benefit doesn't outweigh the issues we have with reconciling it with the way the rest of next works


adgad:
:+1: :tada:

matthew-andrews:
:heart_eyes:

PR #440, opened 2015-12-03
Ordering of modules

ironsidevsquincy:
While we wait for the polyfill service to be fixed in BB10 - https://github.com/3rd-Eden/useragent/issues/83


adgad:
:+1:

PR #444, opened 2015-12-03
Fix overflowing related links on certain breakpoints

adgad:
/cc @ironsidevsquincy 
Before: screen shot 2015-12-03 at 15 04 11

After: 
screen shot 2015-12-03 at 15 04 19

### DIFF
--- a/components/section/main.scss
+++ b/components/section/main.scss
@@ -95,7 +95,7 @@
 
 .column {
 	@include displayFlex;
-	flex-flow: column;
+	flex-direction: column;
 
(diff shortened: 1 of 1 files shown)


components/section/main.scss
```diff
@@ -95,7 +95,7 @@
 
 .column {
 	@include displayFlex;
-	flex-flow: column wrap;
+	flex-flow: column;
```
ironsidevsquincy: could just use `flex-direction:`

ironsidevsquincy:
:+1:

PR #446, opened 2015-12-04
Fix for related content layout jig persisting even without related co…

adgad:
…ntent

/cc @ironsidevsquincy


ironsidevsquincy:
:+1:

PR #452, opened 2015-12-09
Fire initial scroll depth tracking

ironsidevsquincy:
If the page jumps around (e.g. ads), the scroll event is fired, so fire off an initial scroll depth tracking to be safe

cc: @andygout


andygout:
:+1:

PR #453, opened 2015-12-09
Most popular by industry

adgad:
/cc @ironsidevsquincy @matthew-andrews 

* Add dropdown to select most popular by industry
* Refactors the way Homepage is rendered:  (Previously: Layout -> Section -> Cards, Now:  Section -> Cards ). This allows us to only render certain sections client side (and therefore not have to include the entire JSON of the page).
* Reduce DOM size by only rendering dynamic sections (or if the layout tool flag is on) as "react" divs.

TODO:
- [x] Create feature flag
- [x] Put behind feature flag
- [x] Add other industries
- [x] Add "all" to dropdown somehow
- [x] Fix "Warning: render(...): Replacing React-rendered children with a new root component..."
- [x] Styling
- [x] IE9/CORS
- [x] Core experience (maybe just hide)
- [x] Persist choice to localStorage


components/section/browser.js
```diff
@@ -0,0 +1,49 @@
+import SectionNode from './node';
+import nJsonpFetch from 'n-jsonp-fetch';
+import Superstore from 'superstore';
+const fetchres = require('fetchres');
```
ironsidevsquincy: can this be `import`ed too?

components/section/browser.js
```diff
@@ -0,0 +1,49 @@
...
+
+	constructor(props) {
+		super(props);
+		this.defaultSourceStore = new Superstore('local', 'next-frontpage');
+	}
+
+ 	componentWillMount() {
+ 		//If there is a source stored in localStorage, fetch that
+ 		if(this.props.dynamicContent && this.props.dynamicContent.rememberSource) {
+ 			let key = this.props.dynamicContent.rememberSource;
```
ironsidevsquincy: `const`?

components/section/browser.js
adgad: import seems to break fetchres - haven't dug into why though

adgad: actually hold that thought...

components/section/main.scss
```diff
@@ -83,6 +84,18 @@
 	flex-direction: column;
 }
 
+.section-sources__label {
+	font-family: $n-type-sans-serif;
```
ironsidevsquincy: `o-type-...` instead

PR #456, opened 2015-12-11
Amend: Standfirst SCSS to fix size at M layout

andygout:
cc @ironsidevsquincy 

Addresses part of https://github.com/Financial-Times/next-front-page/issues/371 (`change standfirst font size to 18px at M breakpoint (at the moment it's 22px, the same as the headline)`)


ironsidevsquincy:
:+1: 

Guessing we can't deploy anything though, while the a/b test is running

PR #461, opened 2015-12-14
Wording and dropdown changes

adgad:
/cc @tom-parker @ironsidevsquincy 

* Send 'cta change' event (?) when dropdown is changed
* Appends "| alternate-source" to dompath if content is from an alternate source. Presume we don't care which of the sources it is...
* Wording/dropdown content changes


ironsidevsquincy:
:+1:

PR #464, opened 2015-12-17
Fix bizzare BB10 bug where prototype froze

adgad:
@ironsidevsquincy 
*The bug:*
On a BB10 device (with working JS/polyfills), the page would freeze soon after loading, for about a minute, before becoming responsive again.

This was traced to happen on the front page prototype only. After much, much, much deleting and readding of code, it was traced to come from the iJento code - and in the network tab it could be seen that the request to `http://track.ft.com/track/track.js` was stuck pending for 1.2 minutes.

After much, much, much more deleting and readding of code - this was then traced down to the presence of a specific id on the page. The video id.

Why? God knows. Renaming fixes it ¯\_(ツ)_/¯

### DIFF
--- a/components/section/node.js
+++ b/components/section/node.js
@@ -24,6 +24,7 @@ export default class SectionNode extends Component {
 
 	render () {
 		const cols = this.props.cols;
+		const trackable = this.props.trackable || this.props.id;
 		const sectionContentClasses = classify([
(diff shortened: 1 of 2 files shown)


ironsidevsquincy:
:+1:

PR #467, opened 2015-12-21
Only download images if they are going to be shown

adgad:
/cc @andygout @matthew-andrews

Let's be nice to our users.


matthew-andrews:
:smile_cat:

PR #469, opened 2015-12-22
Make IE8 a desktop browser

matthew-andrews:
To align with the rest of the site.

Single apps on their own should not deviate from cross-site standards like this … … … …


adgad:
Does it look/work okay (on both variants)?

matthew-andrews:
my virtual box is broken so i was using brwoserstack to test where the default variant looked quite good actually.  now trying to install virtual box + IE 8 again … may take some time

PR #473, opened 2016-01-04
Output IE8 specific styles in their own stylesheet

adgad:
/cc @ironsidevsquincy @matthew-andrews 

Output IE8 specific grid/card styles in it's own stylesheet for moar performance (results in ~10% less CSS in main.css).

Removes o-grid-container from things that don't go full width (was previously working because .story-card etc css overrides the width, but now comes earlier - but I think o-grid-container should only be on full width things anyway).

Don't merge until https://github.com/Financial-Times/next-build-tools/pull/320 released

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,12 @@ watch-hot-load:
 	node server/dev/init
 
 build:
+	nbt build --main-sass ie8.scss --skip-js --skip-about --skip-haikro --skip-hash --dev
 	nbt build --dev
 
+
 build-production:
+	nbt build --main-sass ie8.scss --skip-js --skip-about --skip-haikro --skip-hash
 	nbt build
 
 smoke:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -35,9 +35,12 @@ watch-hot-load:
 	node server/dev/init
 
 build:
+	nbt build --main-sass ie8.scss --skip-js --skip-about --skip-haikro --skip-hash --dev
```
ironsidevsquincy: As this goes through to `gulp.src` in obt, wonder if you can use a glob here, maybe something like `main*.scss`, and it will compile everything at once (changing the file to `main.ie8.scss`)

adgad: Doesn't seem to work (outputs main*.css with the contents of only main.scss)

ironsidevsquincy:
If it's ok to remove the container classes, :+1:

matthew-andrews:
wow!

PR #475, opened 2016-01-05
Remove unnecessary html

ironsidevsquincy:
* Tidied up the layout config, so there shouldn't be any 'hidden' cards.
 * Made a few assumptions to save on outputting attributes
   * if we never hide a card, omit `data-show`
   * if there is no primaryImage or we never hide the image, omit `data-image-show`
   * if we don't show the image or we never go landscape, omit `data-landscape`
 * Output all the sizes for the `card__title`; makes the css smaller and easier to write
 * Tell the minifier to keep the hidden graduate comment, means we can turn on html minification
 * Remove unused css

Also, as a general tidy, move the sizes into `data-` attributes, rather than classes, to make things more consistent (and smaller?)

### Before
 * HTML - 160Kb
 * CSS - 333Kb

### After (with HTML minification)
 * HTML - 103Kb (36%)
 * CSS - 319Kb (4%)

### DIFF
--- a/components/card/main.scss
+++ b/components/card/main.scss
@@ -27,12 +27,12 @@
 	padding: 0;
 }
 
-[data-card-show~="false"] {
+.card[data-show~="false"] {
 	display: none;
 }
 
-[data-card-landscape~="true"] {
-	&[data-card-has-image~="true"][data-image-show~="true"] {
+.card[data-landscape~="true"] {
+	&[data-has-image~="true"][data-image-show~="true"] {
 		position: relative;
 		min-height: 56px;
 		padding-left: 120px;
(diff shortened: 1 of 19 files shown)


components/card/main.scss
```diff
@@ -44,36 +42,32 @@
 }
 
 @each $layout-name in $_o-grid-layout-names {
-	@media \0screen {
-		[data-card-show~="#{$layout-name}--true"] {
+	@include oGridTargetIE8 {
```
adgad: You'll probably get a merge conflict here - but could you do these in ie8.scss

config/layout.js
```diff
@@ -18,30 +18,30 @@ export default [
...
-				{ size: 'small', landscape: true },
-				{ size: 'small', landscape: true },
-				{ size: 'small', landscape: true },
-				{ size: 'small', landscape: true },
-				{ size: 'small', landscape: true },
-				{ size: 'tiny', landscape: true },
-				{ size: 'tiny', landscape: true },
-				{ size: 'tiny', image: true, landscape: true },
-				{ size: 'tiny', landscape: true }
+				{ column: 0, size: 'large', standFirst: true, image: true, related: 1 },
```
adgad: Not sure what these changes are doing?

adgad:
:+1: (post IE8)

PR #477, opened 2016-01-06
Check for presence of top stories before assuming graphQL response is OK

adgad:

Currently if CAPI is down, graphQL will still return results for videos (with all other keys being null).

This ensures that the poller doesn't overwrite the in-memory data if there are no top stories in the response.


matthew-andrews:
seems good :+1:

PR #478, opened 2016-01-08
Refactor engine

adgad:
engine.js was hard and magical. Rather than handle lots of possible cases which we don't know and aren't use, simplify the cases which are being used and move it to config.

Hopefully this is the first step towards something a lot simpler and easier to iterate on.


ironsidevsquincy:
:+1:

PR #479, opened 2016-01-13
Fix the inline tags for tiny cards

adgad:
Bring back all the horrible loops :(

### DIFF
--- a/components/card/title/main.scss
+++ b/components/card/title/main.scss
@@ -25,7 +25,7 @@
 		margin: 0 5px 0 -6px;
 		background-color: getColor('cold-2');
 	}
-	.card[data-size=tiny] & {
+	.card[data-size~=tiny] & {
 		display: inline;
 
 		&:first-child {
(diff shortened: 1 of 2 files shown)


components/card/title/main.scss
```diff
@@ -109,7 +132,7 @@ $card-title-text-sizes: (
 				@include oGridRespondTo($cardBreakpoint) {
 					@each $text-size, $font-sizes in $text-sizes {
 						.card[data-size~="#{$breakpoint}--#{$text-size}"] .card__title {
-							@include card-title-size($text-size, nth($font-sizes, 1), nth($font-sizes, 2));
+							@include card-title($text-size, nth($font-sizes, 1), nth($font-sizes, 2));
 						}
```
ironsidevsquincy: https://github.com/Financial-Times/next-front-page/blob/71aae150fff9beef0e9d273c152b79402f412f69/components/card/title/main.scss

PR #480, opened 2016-01-14
Myft: anon state

ironsidevsquincy:
Initial anon state for myft (though this probably wont be displayed, it's needed for React to have an initial section to work with on the client)

Design is still in flux, but wanted to get something out now so this PR doesn't get too big (note, it's behind  `frontPageMyftSection` flag)

Also, changed the layout so cards specify the actual component type to use (rather than a string)

screen shot 2016-01-14 at 10 10 15

### DIFF
--- a/components/card/article.js
+++ b/components/card/article.js
@@ -10,10 +10,10 @@ import Related from './related/related';
 
 class Article extends Component {
 	render () {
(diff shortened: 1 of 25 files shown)


adgad:
:+1:

PR #484, opened 2016-01-15
Responsive imgs

andygout:
cc @adgad @ironsidevsquincy

### DIFF
--- a/components/card/image/image.js
+++ b/components/card/image/image.js
@@ -1,19 +1,13 @@
-import qs from 'querystring';
-
 import React, {Component} from 'react';
 import {objMap, responsiveClass} from '../helpers';
 import { Image as NImage } from 'n-image';
-
-const imageOptions = {
-	source: 'next',
-	fit: 'scale-down',
-	compression: 'best'
-};
 
 export default class Image extends Component {
 	render () {
 		const article = this.props.article;
 		const hasHeadshot = article.branding && article.branding.taxonomy === 'authors' && article.branding.headshot;
+		const url = hasHeadshot ? article.branding.headshot : article.primaryImage.rawSrc;
+		const isImgServiceUrl = hasHeadshot ? true : false;
 
 		const srcset = Object.assign({}, this.props.imageSrcSet);
 		if(hasHeadshot) {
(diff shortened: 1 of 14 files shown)


components/card/image/image.js
```diff
@@ -2,32 +2,47 @@ import qs from 'querystring';
...
+	compression: 'best'
 };
 
 export default class Image extends Component {
 	render () {
 		const article = this.props.article;
 		const hasHeadshot = article.branding && article.branding.taxonomy === 'authors' && article.branding.headshot;
-		const imageSrc = hasHeadshot ? `${article.branding.headshot}?${qs.stringify(imageOptions)}` : article.primaryImage.src;
+		const picClass = "card__picture";
+		const imgClass = "card__image";
```
ironsidevsquincy: any reason for these to be stored in a variable, rather than passing the string straight to the react component

config/layouts/editors-picks.js
```diff
@@ -8,7 +8,8 @@ export default [
 				size: { default: 'tiny', S: 'small' },
 				landscape: { default: true, S: false },
 				image: { default: true },
-				imageStick: { default: true}
+				imageStick: { default: true},
+				srcSet: { default: 100, s: 315, m: 277, l: 161, xl: 203 }
```
ironsidevsquincy: probably best to call these `imageSrcSet`, to make it more explicit

components/card/image/main.scss
```diff
@@ -72,11 +72,11 @@
...
-		width: 60%;
-		@include oGridRespondTo(M) {
-			margin-left: auto;
-		}
-	}
+.card__image-link--headshot .card__picture {
+  width: 56.25%;
+  @include oGridRespondTo(M) {
+    margin-left: auto;
+  }
```
ironsidevsquincy: indentation

components/card/image/image.js
andygout: Mainly for readability but it's just as readable passing them in directly so will change.

components/card/image/image.js
```diff
@@ -2,32 +2,47 @@ import qs from 'querystring';
...
+
+		const srcset = Object.assign({}, this.props.srcSet);
+		if(hasHeadshot) {
+			for(var key in srcset) {
```
adgad: This can be a `let`

components/card/image/image.js
```diff
@@ -2,32 +2,47 @@ import qs from 'querystring';
...
+				}
+			}
+		}
+
+		const url =	hasHeadshot
+					?	`${article.branding.headshot}?${qs.stringify(imageOptions)}&width=`
+					:	`https:\/\/next-geebee.ft.com/image/v1/images/raw/` +
+						`${encodeURIComponent(article.primaryImage.rawSrc)}` +
+						`?${qs.stringify(imageOptions)}&width=`
+
```
ironsidevsquincy: indentation

ironsidevsquincy: also, missed this, but to have parity with the n-image handlebars template, url should just be the raw url, and n-image converts to an image-service style url

adgad:
:+1:

PR #488, opened 2016-01-15
Card cleanup

adgad:
@ironsidevsquincy @andygout 

Turns out this has not a huge improvement on performance (-2kb HTML, -9kb CSS), but hopefully a slight improvement in simplicity also


ironsidevsquincy:
:+1:

PR #489, opened 2016-01-18
Fix isomorphic react complaining

adgad:

Various issues:
* Doesn't like the whitespace in the view
* For some reason titles coming back from client graphql query had trailing spaces in them?
* Now need to pass flags to get correct layout


andygout:
:+1:

PR #499, opened 2016-01-21
Delete all the code for the old homepage

adgad:

Delete everything to do with the control homepage and make the variant default.


adgad:
Unminified CSS difference: 259kb vs 395kb

PR #508, opened 2016-01-27
Rows and columns

adgad:
Allow layouts to have any combination of rows and columns.

e.g. (NOTE: local demonstration only, not an actual layout)
screen shot 2016-01-27 at 15 35 03


ironsidevsquincy:
:

PR #509, opened 2016-01-27
React refactor

ironsidevsquincy:
Ostensibly to allow me to update the promo card, we now pass the section, layout and content to the `Section` component, rather than it setting up this data itself

Mainly just a move about of files (split the section config into separate files)


server/routes/front-page.js
```diff
@@ -1,9 +1,8 @@
 import { getData } from '../libs/graphql-poller';
 import sectionContent from '../libs/section-content';
 
-import Feed from '../../components/feed/feed';
```
adgad: Is this being used anywhere else? Can we delete it entirely?

config/layouts/index.js
```diff
@@ -0,0 +1,23 @@
...
+import VideoLayout from './video';
+
+export default {
+    'editors-picks': editorsPicksLayout,
+    'editors-picks-ads': editorsPicksAdsLayout,
+    'featured-section': featuredSectionLayout,
+    'most-popular': mostPopularLayout,
+    'most-popular-ads': mostPopularAdsLayout,
+    myft: myftLayout,
+    'opinion': opinionLayout,
```
adgad: Should be more consistent here (I prefer all in strings, but just taking opinion out of a string would be fine too I guess)

adgad:
:+1:

PR #516, opened 2016-02-01
Add: Temporary fix for Chrome v.48 element 'wiggle' bug

andygout:
cc @ironsidevsquincy @adgad

### DIFF
--- a/components/card/title/main.scss
+++ b/components/card/title/main.scss
@@ -126,6 +126,7 @@ $card-title-text-sizes: (
 	border-bottom: 0;
 
 	&:hover {
-		border-bottom: 1px dotted;
+		text-decoration: underline;
+		text-decoration-style: dotted;
 	}
 }


ironsidevsquincy:
:+1:

keirog:
:cry:

matthew-andrews:
this is arguably a better way to do underlining of text anyway :grin:

adgad:
Interestingly text-decoration-style looks like it's been removed from Chrome Canary
(maybe not so interesting as it's [not actually supported](https://developer.mozilla.org/en/docs/Web/CSS/text-decoration-style)

andygout:
Kept that in so that Firefox users at least have a more consistent experience until Chrome release with fix arrives (I've got a reminder in the diary for when Chrome 49 is released to check if we can revert styling).

adgad:
Ah, I saw it in Chrome 48 with the Experimental Web Features flag on - so probably still is in Canary with that flag on)

PR #519, opened 2016-02-01
Handle markets data response without `change1DayPercent`

ironsidevsquincy:
i.e. don't display commodity

fixes: https://app.getsentry.com/nextftcom/ft-next-frontend/issues/109386134/


adgad:
:+1:

PR #521, opened 2016-02-02
Top Stories Revised: image position

ironsidevsquincy:
left positioned images now sit underneath the tag

screen shot 2016-02-02 at 11 23 23


adgad:
:+1:

PR #522, opened 2016-02-02
Design tweaks and font size changes for new layouts

adgad:
/cc @ironsidevsquincy 

* Move related links closer together
* Remove line after tiny titles
* Font size changes (will affect current layout also, but Simon OK with that)


ironsidevsquincy:
:+1:

PR #524, opened 2016-02-02
Fix IE flex overflow bug

adgad:

### DIFF
--- a/shared/components/column/_column.scss
+++ b/shared/components/column/_column.scss
@@ -3,7 +3,8 @@
     flex-direction: column;
 
     .o-grid-row {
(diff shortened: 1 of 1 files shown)


shared/components/column/_column.scss
```diff
@@ -3,7 +3,8 @@
     flex-direction: column;
 
     .o-grid-row {
-        flex: 1;
+			flex: 1;
+			-ms-flex-preferred-size: auto;
```
ironsidevsquincy: indentation

ironsidevsquincy:
:+1:

PR #525, opened 2016-02-02
Top Story Query: includes primary tag's latest content

ironsidevsquincy:
Adds `topStory` query, which also pulls in the `primaryTag`'s latest three stories, so there is (mostly) always related content

Also removed `top-stories-revised-with-related` layout, as there was no difference

screen shot 2016-02-02 at 15 14 32


shared/components/content/content.js
```diff
@@ -53,7 +53,7 @@ const getData = (item, opts) => {
             data.tag = item.primaryTag;
         }
         if (opts.showRelated) {
-            data.related = item.relatedContent;
+            data.related = item.relatedContent.concat((item.primaryTag && item.primaryTag.items) || []);
```
adgad: Do we need to slice the array here? Where does the limiting to showing 3 happen?

adgad:
:+1:

PR #532, opened 2016-02-03
Fix image position when with related items

ironsidevsquincy:
Also, correct size of image in Opinion section

screen shot 2016-02-03 at 11 36 07


adgad:
:+1:

PR #534, opened 2016-02-03
WIP: Add timestamps to cards

ironsidevsquincy:
(Needs design check)

screen shot 2016-02-03 at 13 26 25

### DIFF
--- a/shared/components/card/timestamp/timestamp.js
+++ b/shared/components/card/timestamp/timestamp.js
@@ -1,0 +1,30 @@
+import ODate from 'o-date';
+import React, { Component } from 'react';
+
+/**
+ * @param {string} date - iso8601
+ */
+export default class extends Component {
+	constructor(props) {
+		super(props);
+		this.state = {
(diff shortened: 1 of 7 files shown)


shared/components/card/timestamp/timestamp.js
```diff
@@ -0,0 +1,30 @@
...
+		this.state = {
+			dateText: ODate.format(props.date, 'date'),
+			title: ODate.format(props.date, 'datetime')
+		};
+	}
+
+	// not called on the initial render, e.g. client only code
+	componentWillReceiveProps(props) {
+		this.setState({
+			dateText: ODate.ftTime(new Date(props.date))
```
adgad: By doing it like this (and not oDate.init) doesn't that mean the time won't update if I leave the page open?

ironsidevsquincy: I *think* it will update, because there's still a reference to the element from the original `ODate.init()` call

PR #535, opened 2016-02-03
Ie10 image overflow

adgad:
* FIxes image overflowing out of their container in IE10
* Couple of IE8 layout fixes
* Workaround for the Chrome 48 jumpy bug (and bring back dotted underline)

@ironsidevsquincy


ironsidevsquincy:
:+1:

PR #536, opened 2016-02-04
If no content, dont render section

adgad:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ verify:
 	nbt verify --skip-layout-checks
 
 unit-test:
-	export NODE_ENV=test; mocha  --require test/setup --recursive --reporter spec test/components
+	export NODE_ENV=test; mocha  --require test/setup --recursive --reporter spec test/server
 
 test: verify unit-test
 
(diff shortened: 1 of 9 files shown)


matthew-andrews:
this is fine for most sections but i _think_ the top stories section should always try render and trigger 5xx (and hope that Fastly will serve stale instead)…  is that compatible with this change?

adgad:
[This](https://github.com/Financial-Times/next-front-page/blob/master/server/routes/front-page.js#L8) should have been - but the property has been renamed since :( 

I will fix and write a test.

andygout:
More for my education, but what is difference between:-

```js
if(!(this.state.content && this.state.content.main && this.state.content.main.length)) {
	return null;
 }
```

and

```js
if(!this.state.content.main.length) {
	return null;
 }
```
?  On the surface it seems this single statement queries the desired status?

adgad:
If `this.state.content` is null, then trying to access `this.state.content.main` will throw an error.

andygout:
And it'll check each one in sequence so will only hit a null rather than an error - got it, thanks!

adgad:
(The property hadn't been renamed - i misread. But added the other first top story to that check as well)

PR #538, opened 2016-02-05
Ability to have image top/bottom/left/right

ironsidevsquincy:
Horrible one this, but the image is now rendered after the tag. This means it only involves a bit of floating to align it left or right, and absolute positioning to put it at the top or the bottom


shared/components/card/standfirst/_standfirst.scss
```diff
@@ -4,16 +4,12 @@
...
 	margin: 5px 0 0;
+	clear: left;
 
 	.card[data-size^="large"] & {
-		font-size: 22px;
-		line-height: 24px;
-	}
-}
-
-@include oGridRespondTo($until: L) {
```
adgad: Don't we need this?

adgad:
:+1: (besides comment)

shared/components/card/standfirst/_standfirst.scss
ironsidevsquincy: Don't think so, logic seemed off previously (i.e. we we're saying `large` was the same as normal up to `L`

https://github.com/Financial-Times/next-front-page/blob/master/shared/components/card/standfirst/_standfirst.scss

PR #540, opened 2016-02-08
Non-obt verify

ironsidevsquincy:
Looks like `lintspaces` wasn't being run, hence big PR

Main update is to the `Makefile` - https://github.com/Financial-Times/next-front-page/compare/quicker-verify?expand=1#diff-b67911656ef5d18c4ae36cb6741b7965R6


adgad:
:+1:

PR #546, opened 2016-02-10
N Makefile

matthew-andrews:


n.Makefile
```diff
@@ -0,0 +1,101 @@
...
+	@git clean -fxd
+	@echo $(DONE)
+
+# install
+instal%: node_modules bower_components _install_scss_lint
+	@for n in $(shell ls functions 2>/dev/null); do $(MAKE) functions/$$n/node_modules; done
+	@echo $(DONE)
+
+# deploy
+deplo%:
```
adgad: Out of interest, what is the % here?

adgad:
:+1:

n.Makefile
matthew-andrews: for other's benefit, it's to help with inheritance and overriding of tasks

n.Makefile
```diff
@@ -0,0 +1,101 @@
...
+#
+# Two main types of tasks.  Common and sub.
+#
+# Common tasks are the public facing API of the Makefile.  Developers/CI should run these commands.
+# Sub-tasks are the inner, private workings of n-makefile.  Developers/CI ought not to run these commands.
+#
+# Type specific conventions
+#
+# - For ‘common’ tasks
+#     - name must match the directory name they generate
```
ironsidevsquincy: does this mean `make build` should create a `build` dir, rather than `public`?

n.Makefile
```diff
@@ -0,0 +1,101 @@
...
+#
+
+# clean
+clea%:
+	@git clean -fxd
+	@echo $(DONE)
+
+# install
+instal%: node_modules bower_components _install_scss_lint
+	@for n in $(shell ls functions 2>/dev/null); do $(MAKE) functions/$$n/node_modules; done
```
ironsidevsquincy: what does this line do?

n.Makefile
matthew-andrews: ah no, this is for sub-tasks not common ones:-
https://github.com/Financial-Times/n-makefile/blob/master/Makefile#L21-L22

i think maybe make build could be a special case but we could do something like:-

```
build:
	@for n in $(shell ls client/[^_]*.scss 2>/dev/null); do $(MAKE) public/$$n; done
	@for n in $(shell ls client/*.js 2>/dev/null); do $(MAKE) public/$$n; done

public/%.scss:
	obt build --scssFile $@ blah blah blah

public/%.js:
	obt build --jsFile $@ blah blah blah
```

ironsidevsquincy:
:+1:

PR #550, opened 2016-02-11
Most popular size fix

adgad:
/cc @andygout @ironsidevsquincy 

The huge most popular card was because that card shows and hides it's image on different breakpoints. 

Rather than try and fix that CSS (with the padding) which would be massively complicated, just duplicate the card for now, one with an image and one without. This means we can get rid of the image.show property.

Hopefully the most popular section simplifies itself soon, so we can ditch all those complications.


andygout:
:+1:

PR #552, opened 2016-02-11
Amend: Srcset values for top stories landscape layout

andygout:
cc @adgad


ironsidevsquincy:
I wonder if we can be cleverer and calculate the sizes automatically based on colspan/image position? might get insanely complicated though

adgad:
Yeah it's a bit of a pain updating at the moment. 

Simon is talking about only having 3/4 different image sizes on the page, so if that works out maybe we could just have a set of 4 different srcSets ( imageSizes.large, imageSizes.small etc)

PR #557, opened 2016-02-12
Amend: Top stories landscape layout (stories 2-5 match picture story layout)

andygout:
cc @adgad @ironsidevsquincy 

N.B. Story 5 is current a story without a picture, but will display image same as story 2.

screen shot 2016-02-12 at 15 45 22

### DIFF
--- a/config/layouts/top-stories-landscape.js
+++ b/config/layouts/top-stories-landscape.js
@@ -59,24 +59,26 @@ export default [
 								components: [
 									{
 										type: Content,
-										size: 'medium',
-										showStandfirst: true
+										size: 'medium'
+									},
+									{
+										type: Content,
+										size: 'medium'
 									}
 								]
 							},
-								//Column 1
+							//Column 2
 							{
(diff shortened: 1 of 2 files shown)


config/layouts/top-stories-landscape.js
```diff
@@ -59,24 +59,26 @@ export default [
...
 										type: Content,
-										size: 'medium'
-									},
-									{
-										type: Content,
-										size: 'medium'
-
+										size: 'medium',
+										image: {
+											sizes: { default: 100, s: 100, m: 277, l: 226, xl: 233 },
```
adgad: don't think you need to specify s here, since it's the same as default

adgad:
:+1:

config/layouts/top-stories-landscape.js
andygout: Will sort in separate pull request as this also happens in `top-stories-picture-story`.

PR #559, opened 2016-02-12
Make third column of top stories (default) bigger

adgad:
Before:
screen shot 2016-02-12 at 16 26 33

After:
screen shot 2016-02-12 at 16 26 27


ironsidevsquincy:
:+1:

PR #563, opened 2016-02-15
Pass responsibility to managing make verify to n-makefile, …

matthew-andrews:
and also pass responsibility for managing the .editorconfig and .scss-lint.yml files to n-makefile


.gitignore
```diff
@@ -12,3 +12,5 @@
 /test/browser/reports/
 
 !/nodemon.json
+.editorconfig
+.scss-lint.yml
```
ironsidevsquincy: and `.eslintrc.json`

matthew-andrews: right now that doesn't have the required react plugins.

i will centralise this but I'm going to have to do this in a more intelligent way

ironsidevsquincy:
:+1:

PR #564, opened 2016-02-15
Switch to core/enhanced selectors

keirog:


ironsidevsquincy:
:+1: 

Though this reminds me, we should be using `n-util` for our core/enhanced hiding/showing needs

keirog:
Ah, yep

PR #575, opened 2016-02-17
Remove the second two MPUs and all the complexity around them

adgad:
/cc @ironsidevsquincy @VladDubrovskis @andrewgeorgiou1981 @andygout 
* Remove most popular ads layout and editors picks ads layout
* Switch to the 'default' ad layout in next-ads-components
* Editors picks sticking with 4 stories for now (need to verify this is okay)

screen shot 2016-02-17 at 14 40 19


ironsidevsquincy:
Awesome :+1:

PR #586, opened 2016-02-24
Track scroll depth

andygout:
cc @ironsidevsquincy @adambraimbridge 

For scroll depth graphs it seems the only data needed is the `data-trackable` value of the sections (i.e. `opinion`, `editors-picks`, etc.) so have dispensed with sending results from `getDomPath` function (which do not seem to apply in any case given the sections have no ancestors with a `data-trackable` property).

Will continue to send `componentPos` should sections ever be re-ordered and we need to know when, e.g. opinion is at a given position in the ordering of sections.

`js-track-scroll-event` property is added to the parent div of the section so that we can easily acquire the `id` value; in the case of the top-stories variations, this will allow us to consistently track the section as `top-stories` (rather than, e.g. `top-stories-picture-story`).


ironsidevsquincy:
:+1:

PR #587, opened 2016-02-24
No need for a flag, just use query param

adgad:
Will make it easier to demo

@ironsidevsquincy @andygout @VladDubrovskis


VladDubrovskis:
:+1:

PR #595, opened 2016-02-26
Fix: IE8 hidden x-overflow area bug

andygout:
cc @ironsidevsquincy 

Fixes: https://jira.ft.com/browse/NFT-336 (below is illustration of the issue)

screen shot 2016-02-15 at 17 40 49


ironsidevsquincy:
:+1:

PR #598, opened 2016-02-29
Amend: FastFT to act as link (not toggle) for layouts default-M

andygout:
cc @ironsidevsquincy

### DIFF
--- a/shared/components/section/_section.scss
+++ b/shared/components/section/_section.scss
@@ -72,6 +72,12 @@
 	flex-direction: column;
 }
 
+.section__column--hide-until-l {
+	@include oGridRespondTo($until: L) {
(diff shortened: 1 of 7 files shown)


shared/components/section/_section.scss
```diff
@@ -72,6 +72,12 @@
 	flex-direction: column;
 }
 
+.section__column--hide-until-l {
```
ironsidevsquincy: capitalise `l`

ironsidevsquincy:
:+1:

shared/components/section/_section.scss
andygout: scss linting rules only allow lower case letters in class names.

PR #600, opened 2016-03-01
Fix: Top Stories link path when FastFT Journey flag switched on

andygout:
cc @ironsidevsquincy 

Fixes issue with `frontPageFastFTJourney` flag switched on of 'Top Stories' link linking down to Top Stories section via fragment identifier; is now a link back to itself (home page).


andygout:
Am now including this change in separate branch: https://github.com/Financial-Times/next-front-page/tree/style-header-tabs-for-fastft-journey

PR #609, opened 2016-03-07
Remove: Unwanted block in subnav

andygout:
cc @ironsidevsquincy 

Fixes this [issue](https://jira.ft.com/browse/NFT-352) in short-term.

However, should the separating border have a gap between the top of the cards and where the border starts, as it does at the bottom of the cards (see images)? If so, the issue could potentially be fixed without amending the z-indexes in `_column.scss` by adding 15px to the border's [top position](https://github.com/Financial-Times/next-front-page/blob/master/shared/components/column/_column.scss#L18) (although not sure how to do that in a (non-hacky) way that adheres to `oGridGutter` logic).

Long-term solutions:-
- Applying margin-top to entire column rather than the individual cards (but I remember us doing this for one of the Top Stories layouts and having to abandon for a reason I now can't remember, and it is quite an overhaul).

- @keirog has also started some z-index logic management in [`next-sass-setup`](https://github.com/Financial-Times/next-sass-setup/blob/master/_z-indexes.scss)

But will this do it in the meantime?

##### Top of cards
screen shot 2016-03-07 at 14 32 23

##### Bottom of cards
screen shot 2016-03-07 at 14 32 41

### DIFF
--- a/shared/components/column/_column.scss
+++ b/shared/components/column/_column.scss
@@ -11,7 +11,7 @@
 		top: oGridGutter() / 2;
 		bottom: oGridGutter() / 2;
 		background-color: oColorsGetColorFor(column, border);
-		z-index: 1;
+		z-index: 0;
 
 		@include oGridRespondTo(M) {
 			left: floor(oGridGutter(M) / 2);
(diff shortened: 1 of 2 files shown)


ironsidevsquincy:
Possibly, though then the top of the lines wont line up with the horizontal borders

screen shot 2016-03-07 at 15 05 34

Borders might well change in the future.

If reducing the z-indices has been tested in different scenarios (menus open, etc), then :+1:

PR #619, opened 2016-03-10
Remove: Most Popular By Industry

andygout:
cc @ironsidevsquincy


config/queries.js
```diff
@@ -187,26 +186,10 @@ const frontPage = (region) => (`
...
-	query PopularArticlesBy {
-		mostPopular: popularFromHui(${facet}: "${uuid}") {
-			... Basic
-			... Extended
-			... OpinionData
-		}
-	}
-`);
-
 const popularArticles = `
```
ironsidevsquincy: is this still needed?

config/queries.js
```diff
@@ -68,7 +68,6 @@ const frontPage = (region) => (`
 	${fragments.related}
 	${fragments.opinionData}
 	${fragments.liveBlog}
-
```
ironsidevsquincy: intentional?

shared/components/section/sources/_sources.scss
```diff
@@ -1,15 +0,0 @@
...
-
-.section-sources__label {
-	font-family: $o-typography-sans;
-	font-weight: oFontsWeight(light);
-	padding-right: 8px;
-}
-
-.section-sources__select {
-	@include oFormsCommonField();
-	@include oFormsSelect();
```
ironsidevsquincy: wonder if we still depend on `o-forms` now?

ironsidevsquincy:
sweet :+1:

shared/components/section/sources/_sources.scss
andygout: Only dependencies use it - nothing directly in the frontpage, so will remove.

PR #634, opened 2016-03-15
Add: aria-label for markets data figures

andygout:
cc @ironsidevsquincy

### DIFF
--- a/client/components/markets-data/markets-data.js
+++ b/client/components/markets-data/markets-data.js
@@ -67,9 +67,8 @@ const regionalSecurities = {
 	]
 };
 
-const createAriaLabel = (priceChangeDirection, percentChange) => {
-	return priceChangeDirection + ' ' + percentChange.toString().replace(/\./, ' point ') + '%';
-};
+const createAriaLabel = (priceChangeDirection, percentChange) =>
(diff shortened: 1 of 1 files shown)


client/components/markets-data/markets-data.js
```diff
@@ -67,6 +67,10 @@ const regionalSecurities = {
 	]
 };
 
+const createAriaLabel = (priceChangeDirection, percentChange) => {
+	return priceChangeDirection + ' ' + percentChange.toString().replace(/\./, ' point ') + '%';
```
ironsidevsquincy: can use string interpolation

also, is the `%` symbol read out?

client/components/markets-data/markets-data.js
```diff
@@ -107,7 +111,8 @@ export default flags => {
 								class="markets-data__item__link"
 								data-trackable="link">
 								<h2 class="markets-data__item__name">${security.name}</h2>
-								<p class="markets-data__item__change markets-data__item__change--${priceChangeDirection}">
+								<p 	class="markets-data__item__change markets-data__item__change--${priceChangeDirection}"
```
ironsidevsquincy: odd indentation

PR #635, opened 2016-03-15
Add: aria-label for myFT promo box link

andygout:
cc @ironsidevsquincy 

Screen reader now says "myFT" rather than "my featuring".


ironsidevsquincy:
Crazy :+1:

PR #636, opened 2016-03-16
Add: aria-label for fastFT link

andygout:
cc @ironsidevsquincy 

Screen reader now says "fastFT" rather than "fast featuring".


ironsidevsquincy:
:+1:

PR #637, opened 2016-03-16
Set: Landmark roles for screen reader

andygout:
cc @ironsidevsquincy 

Top Stories section (along with fastFT) now listed as 'main', with subsequent sections as 'complementary'.

fastFT sidebar exists as an aside (default ARIA role value is 'complementary') within the 'main' section, which seems to be acceptable providing that the aside relates to the 'main', which in this context feels it does: http://stackoverflow.com/questions/17549780/can-you-use-an-aside-element-inside-a-section-role-main-element

[W3C - The Roles Model: `main`](https://www.w3.org/TR/wai-aria/roles#main)
[W3C - The Roles Model: `complementary`](https://www.w3.org/TR/wai-aria/roles#complementary)

### DIFF
--- a/views/front-page.html
+++ b/views/front-page.html
@@ -46,11 +46,10 @@
 	-->
 	<div class="ad-placeholder ad-placeholder--below-header"></div>
 
-	<div class="o-grid-row">
+	<div role="main" class="o-grid-row">
 	{{#each sections}}
 		<div
 			id="{{id}}"
-			{{#if role}}role="{{role}}"{{/if}}
 			{{#if trackScrollEvent}} class="js-track-scroll-event"{{/if}}
 			data-o-grid-colspan="{{colspan size}}">{{#unless isClient}}{{{reactRenderToString ../Section spread=@this}}}{{/unless}}</div>
 	{{/each}}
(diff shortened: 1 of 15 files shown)


views/front-page.html
```diff
@@ -50,8 +50,10 @@
 	{{#each sections}}
 		<div
 			id="{{id}}"
+			{{#if role}}role="{{role}}"{{/if}}
 			{{#if trackScrollEvent}} class="js-track-scroll-event"{{/if}}
-			data-o-grid-colspan="{{colspan size}}">{{#unless isClient}}{{{reactRenderToString ../Section spread=@this}}}{{/unless}}</div>
+			data-o-grid-colspan="{{colspan size}}">{{#unless isClient}}{{{reactRenderToString ../Section spread=@this}}}{{/unless}}
+		</div>
```
ironsidevsquincy: think the closing div needs to be on the same line, or else react complains about empty spaces

ironsidevsquincy:
Is the fast ft block related to the top stories?

andygout:
Following discussion, will contain all sections with `role="main"` and address any elements contained therein that default to `role="complementary"` (disabling that role value).

andygout:
Given I have changed the `aside` elements to `div`s, I have also amended the fastFT section's associated values from 'aside' to 'sidebar' given it is referred to as such (or 'sidebarComponent').

ironsidevsquincy:
:+1:

PR #643, opened 2016-03-18
Amend: myFT promo card logic

andygout:
cc @ironsidevsquincy @phamann @keirog @laurieboyes 

Promo card will no longer be displayed when following two or more topics.

##### Following 0 topics
following 0

##### Following 1 topic
following 1

##### Following 2 topics
following 2

##### Following 3 topics
following 3

##### Following 4 topics
following 4

### DIFF
--- a/client/components/myft/myft.js
+++ b/client/components/myft/myft.js
@@ -35,8 +35,8 @@ const filterDuplicateArticles = (articles, concept) => {
 	return articles.concat(concept.items);
 };
 
-const mapCards = (length, card) => {
-	return length > 1 ? Object.assign(card, { type: Content }) : Object.assign(card, { isMyftUser: true })
+const assignAttributes = (promoBoxRequired, card) => {
+	promoBoxRequired ? Object.assign(card, { isMyftUser: true }) : Object.assign(card, { type: Content })
 }
 
 const handleResponse = (myFtContainerEl, myftClient, flags, response) => {
@@ -57,7 +57,7 @@ const handleResponse = (myFtContainerEl, myftClient, flags, response) => {
 			col.components
 				.reduce((prev, column) => prev.concat(column.components), [])
 				.filter(card => card.type === MyftPromo)
-				.map(mapCards.bind(null, followed.length))
+				.map(assignAttributes.bind(null, followed.length < 2))
 		);
 	}
 
--- a/client/components/myft/myft.js
+++ b/client/components/myft/myft.js
@@ -44,6 +44,12 @@ const convertToContent = (shouldConvert, card) => {
 	}
 }
 
+// recursively seek innermost components and call conversion function
+const changeComponent = (shouldConvert, component) =>
+	component.components ?
+		Object.assign({}, component, { components: component.components.map(changeComponent.bind(null, shouldConvert)) }) :
+		convertToContent(shouldConvert, component);
+
 const handleResponse = (myFtContainerEl, myftClient, flags, response) => {
 	const { popularTopics, user: { viewed, followed }} = response;
 	// flag up followed concepts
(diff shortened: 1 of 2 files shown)


config/queries.js
```diff
@@ -220,7 +220,7 @@ const user = `
 			viewed(limit: 3) {
 				... Basic
 			}
-			followed(limit: 3) {
+			followed(limit: 4) {
```
ironsidevsquincy: don't all `limit`s in this query need to be 4?

client/components/myft/myft.js
```diff
@@ -34,6 +35,10 @@ const filterDuplicateArticles = (articles, concept) => {
 	return articles.concat(concept.items);
 };
 
+const mapCards = (length, card) => {
+	return length > 1 ? Object.assign(card, { type: Content }) : Object.assign(card, { isMyftUser: true })
```
ironsidevsquincy: single line, no need for return

client/components/myft/myft.js
```diff
@@ -34,6 +35,10 @@ const filterDuplicateArticles = (articles, concept) => {
 	return articles.concat(concept.items);
 };
 
+const mapCards = (length, card) => {
```
ironsidevsquincy: better name than `mapCards`? What is this function doing?

client/components/myft/myft.js
ironsidevsquincy: Also, unnecessary optimisation, but length is a fixed value, so instead of testing `length > 1` each time, why not pass in a boolean and do the test once when binding, e.g. `mapCards.bind(null, followed.length > 1)`

config/queries.js
andygout: The only scenario in which four cards of exclusively one type is needed is when you are following 4+ topics and all four slots are occupied by 'followed' content.

The maximum number of 'popularTopics' or 'viewed' is three for each respectively (depending on whether you have viewed 0 topics or 3+ topics) as the promo box will always occupy the first of the four spaces when all cards are exclusively 'popularTopics' or exclusively 'viewed'.

client/components/myft/myft.js
```diff
@@ -34,6 +35,10 @@ const filterDuplicateArticles = (articles, concept) => {
 	return articles.concat(concept.items);
 };
 
+const assignAttributes = (promoBoxRequired, card) => {
+	promoBoxRequired ? Object.assign(card, { isMyftUser: true }) : Object.assign(card, { type: Content })
```
ironsidevsquincy: not the correct syntax for the 'concise body' version - https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Function_body

client/components/myft/myft.js
```diff
@@ -34,6 +35,10 @@ const filterDuplicateArticles = (articles, concept) => {
...
+}
```
ironsidevsquincy: not something to worry about really, just being picky, but the names of the function/arguments are a bit meaningless. think of this function as a self-contained thing, seen in isolation. what does `promoBoxRequired` mean? seems to imply some knowledge of a promo being shown elsewhere. what does this function do? converts a card to a `Content` type (or not). So maybe `convertToContent` as function name, and `shouldCovert` as argument?

config/queries.js
ironsidevsquincy: It's more for deduping though. Say you're following 2 topics, and only getting back 3 viewed and 3 popular topics (which are equal), both of which have the followed topics in them. Then after deduping you'll only have 3 topics to show in the section (as opposed to 4)

client/components/myft/myft.js
```diff
@@ -42,20 +52,20 @@ const handleResponse = (myFtContainerEl, myftClient, flags, response) => {
...
-			col.components
-				.reduce((prev, column) => prev.concat(column.components), [])
-				.filter(card => card.type === MyftPromo)
-				.map(card => Object.assign(card, { isMyftUser: true })
-			)
-		);
-	}
+
+	// recursively seek innermost components and call conversion function
+	const changeComponent = component =>
```
ironsidevsquincy: doesn't need to live within this function

ironsidevsquincy:
bar comment, :+1:

PR #644, opened 2016-03-18
Sample scrolldepth events

tom-parker:
cc @andygout @ironsidevsquincy @commuterjoy 

I'm hoping this is ok to sample the scroll depth events from the frontpage. We saw ~400K of these from the front page yesterday which is almost a third of our daily allowance.

### DIFF
--- a/client/components/scroll-depth/scroll-depth.js
+++ b/client/components/scroll-depth/scroll-depth.js
@@ -22,10 +22,10 @@ const scrollHandlerFactory = () => {
 };
 
 export default () => {
-	const scrollHandler = scrollHandlerFactory();
 	if (!/spoor-id=0/.test(document.cookie)) {
 		return false;
 	}
+	const scrollHandler = scrollHandlerFactory();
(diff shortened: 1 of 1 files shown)


andygout:
What proportion of the existing scroll depth events will we continue to catch - 10% (I think I remember being mentioned)?  I'd imagine that 40K per day would still be sufficient data.

tom-parker:
Think it works out ~6.5%

client/components/scroll-depth/scroll-depth.js
```diff
@@ -23,6 +23,9 @@ const scrollHandlerFactory = () => {
 
 export default () => {
 	const scrollHandler = scrollHandlerFactory();
+	if (!/spoor-id=0/.test(document.cookie)) {
```
ironsidevsquincy: any reason not to put this above the above line?

tom-parker: Nope!

ironsidevsquincy:
:+1:

PR #648, opened 2016-03-21
Test element visible

adgad:

For, er, obvious reasons


matthew-andrews:
can we test that this will fail if we `display: none;` the page?

adgad:
I have indeed tested that case locally: https://saucelabs.com/tests/06620272bbc541d0bd34af78aef4664a

matthew-andrews:
+!

PR #649, opened 2016-03-22
Etching out of what n-makefile.json could be

matthew-andrews:
- Combines the two webpack config files.
- Removes the requirement for people to have `node_modules/.bin` on their paths.
- Defines the entry point for Sass + JS in a config file.

@ironsidevsquincy


webpack.config.js
```diff
@@ -3,15 +3,12 @@ const autoprefixer = require('autoprefixer');
 const BowerWebpackPlugin = require('bower-webpack-plugin');
```
matthew-andrews: I'll make a second PR later that will pull this file out and put it into the `n-makefile` and then try it out in another app :smile:

webpack.config.js
```diff
@@ -30,7 +23,7 @@ module.exports = {
 				],
 				query: {
 					cacheDirectory: true,
-					presets: ['react', 'es2015'],
+					presets: (require('./package.json').dependencies.react ? ['react', 'es2015'] : ['es2015']),
```
matthew-andrews: for other apps mostly

webpack.config.js
```diff
@@ -3,19 +3,12 @@ const autoprefixer = require('autoprefixer');
...
-	entry: {
-		'main.js': './main.js',
-		'main.css': './main.scss',
-		'ie8.css': './ie8.scss'
-	},
-	output: {
-		path: './public',
-		filename: '[name]'
-	},
+	devtool: process.argv.indexOf('--dev') === -1 ? 'source-map' : 'cheap-module-eval-source-map',
```
ironsidevsquincy: we've had a few issues with using `cheap-module-eval-source-map`, i.e. breakpoints don't work probably. might be worth just using `source-map` for all - though will have a performance impact. or maybe one of the other options

https://webpack.github.io/docs/configuration.html#devtool

webpack.config.js
```diff
@@ -3,15 +3,12 @@ const autoprefixer = require('autoprefixer');
...
 module.exports = {
-	devtool: 'source-map',
+	devtool: process.argv.indexOf('--dev') === -1 ? 'source-map' : 'cheap-module-eval-source-map',
 	context: path.join(__dirname, 'client'),
-	entry: {
-		'main.js': './main.js',
-		'main.css': './main.scss',
-		'ie8.css': './ie8.scss'
-	},
+	entry: config.assets.entry,
```
ironsidevsquincy: so simple - that's cool

ironsidevsquincy:
:+1:

webpack.config.js
matthew-andrews: ok, i'll do that

PR #650, opened 2016-03-22
Allow bower/node module linking

ironsidevsquincy:
Odd issue in babel 6 where is borks when linking modules

This fixes it (see http://stackoverflow.com/questions/34574403/how-to-set-resolve-for-babel-loader-presets/)

@adgad


adgad:
(╯°□°)╯︵ ┻━┻ 

:+1:

matthew-andrews:
ahhh…  https://github.com/Financial-Times/n-makefile/blob/master/config/webpack.config.js

PR #662, opened 2016-03-24
Don't show tag/brand on live blog cards

ironsidevsquincy:


nodemon.json
```diff
@@ -1,4 +1,5 @@
 {
+  "ignoreRoot": [],
```
ironsidevsquincy: so nodemon watch `node_modules`

PR #663, opened 2016-03-24
Stops babel sending polyfill's

GlynnPhillips:
We use the FT polyfill service so using the babel polyfill's causes
duplication.

I have tested one various devices and and browsers and haven't seen anything break or console errors. I have also grep'ed the code base for use of stuff like `symbol` and came across nothing but there might be some I have missed.

Would be nice to get some testing / feedback with people who know the product better maybe @ironsidevsquincy or @andygout ?

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -30,7 +30,7 @@ module.exports = {
 					),
 					plugins: [
 						require.resolve('babel-plugin-add-module-exports'),
-						require.resolve('babel-plugin-transform-runtime'),
+						[require.resolve('babel-plugin-transform-runtime'), { polyfill: false }],
 						[require.resolve('babel-plugin-transform-es2015-classes'), { loose: true }]
 					]
 				}


matthew-andrews:
can you tweak this in here too? :)

https://github.com/Financial-Times/n-makefile/blob/master/config/webpack.config.js

ironsidevsquincy:
Any idea how many kb this saves?

GlynnPhillips:
@matthew-andrews Done https://github.com/Financial-Times/n-makefile/pull/27 (wasn't sure about testing this)

@ironsidevsquincy Its ~50kb and I guess thats more significant on lower-end devices and for storing in memory and parsing

ironsidevsquincy:
:+1: 

Keep an eye on the frontend errors

https://app.getsentry.com/nextftcom/ft-next-frontend/?query=is%3Aunresolved%20appName%3Afront-page%20

PR #669, opened 2016-03-29
Adds some accessibility improvements

GlynnPhillips:
- Level 1 heading (visually hidden)
- Updating `id` so skip to content link works
- Adds level 2 heading for top stories (visually hidden)

### DIFF
--- a/shared/components/section/section.js
+++ b/shared/components/section/section.js
@@ -38,7 +38,7 @@ export default class extends Component {
 				{
 					cols.meta ?
 						<div data-o-grid-colspan={colspan(cols.meta)} className="section__column section__column--meta">
-							<SectionMeta title={this.props.title} />
+							<SectionMeta title={this.props.title} meta={this.props.cols.meta} />
 						</div> :
 						null
 				}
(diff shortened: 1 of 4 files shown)


views/front-page.html
```diff
@@ -13,8 +13,10 @@
 {{/if}}
 
 <section
-	id="main-body"
+	id="site-content"
```
ironsidevsquincy: why this change

shared/components/section/section.js
```diff
@@ -38,7 +38,7 @@ export default class extends Component {
 				{
 					cols.meta ?
 						<div data-o-grid-colspan={colspan(cols.meta)} className="section__column section__column--meta">
-							<SectionMeta title={this.props.title} />
+							<SectionMeta title={this.props.title} meta={this.props.cols.meta} />
```
ironsidevsquincy: I would only pass in the things you need, in this case

```... isHidden={this.props.cols.meta.hide} ...```

views/front-page.html
GlynnPhillips: This makes the skip links work, example where its failing on the homepage: https://next.ft.com/uk#site-content

and working on an article: https://next.ft.com/content/57981914-f588-11e5-96db-fc683b5e52db#site-content

I looked around and couldn't see the `id` being used else where

shared/components/section/section.js
GlynnPhillips: Sure, I will update this.

ironsidevsquincy:
:+1:

PR #674, opened 2016-03-30
Section Header Tweaks

ironsidevsquincy:
* links to sections opinion and video sections
 * hover state for myFT section header

@adgad


adgad:
wow that was quick!

PR #678, opened 2016-03-30
Opinion layout

andygout:
cc @ironsidevsquincy 

##### L - XXL
screen shot 2016-03-30 at 14 41 38

##### M
screen shot 2016-03-30 at 14 42 01

##### default - S
screen shot 2016-03-30 at 14 43 10


ironsidevsquincy:
:+1:

Issue #33, opened 2015-06-30
FastFT logo on Firefox looks silly at all resolutions


a-rhombus:
![fastft-logo-firefox](https://cloud.githubusercontent.com/assets/13115898/8428912/b45c34c2-1f1a-11e5-8ae9-487f02079551.png)

PR #684, opened 2016-04-01
Update n-makefile.json

phamann:
As per [n-makefile #33](https://github.com/Financial-Times/n-makefile/pull/33) app specific include paths are now defined via the config file.


matthew-andrews:
covered in another pr

PR #689, opened 2016-04-01
Featured section layout (homepage audit)

andygout:
cc @ironsidevsquincy 

N.B. Before going live we must first speak to Editorial to ensure that at least four stories in the 'analysis' or 'comment' genre are available (per section: Technology/Markets News/Life & Arts), as there have been instances in the dev process in which there are only two (the current requirement).

##### default (Technology)
default tech

##### default (Markets News)
default markets

##### S
s

##### M - L
m-l

##### XL - XXL
xl-xxl

### DIFF
--- a/config/layouts/featured-section-new.js
+++ b/config/layouts/featured-section-new.js
@@ -26,7 +26,7 @@ export default [
 				components: [
 					{
 						type: Content,
-						size: 'small',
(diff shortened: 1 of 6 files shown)


ironsidevsquincy:
:+1:

PR #692, opened 2016-04-04
Revert: Highlights section to previous layout (when titled Editor's Picks)

andygout:
cc @ironsidevsquincy


ironsidevsquincy:
:+1:

PR #693, opened 2016-04-04
Amend: Opinion section default-M layout - all stories full column width

andygout:
cc @ironsidevsquincy 

##### From:
from

##### To:
to


ironsidevsquincy:
:+1:

PR #694, opened 2016-04-04
Pass responsibility to n.Makefile for compressing JS, hashing assets and tar'ing apps

matthew-andrews:


package.json
```diff
@@ -45,6 +45,7 @@
     "eslint": "^1.10.3",
     "eslint-plugin-react": "^3.16.1",
     "extract-text-webpack-plugin": "^1.0.1",
+    "haikro": "^2.3.2",
```
ironsidevsquincy: do we need this now?

matthew-andrews: `build`+`build-production` is now completely NBT free so yes, for now, want to tidy up the dependencies but will do that in a later version :]

Makefile
```diff
@@ -27,8 +27,6 @@ build: _webpack_setup
 
 build-production: _webpack_setup
```
matthew-andrews: is the `_webpack_setup` task still needed? @ironsidevsquincy

ironsidevsquincy: Says the issue's resolved, will check

n-makefile.json
```diff
@@ -4,6 +4,9 @@
       "./public/main.js": "./client/main.js",
       "./public/main.css": "./client/main.scss",
       "./public/ie8.css": "./client/ie8.scss"
-    }
+    },
+    "includes": [
+      "/n-card/"
+    ]
```
ironsidevsquincy: don't need this after doing a rebase from master

ironsidevsquincy:
:+1:, though had to get the latest n-makefile already, so do a rebase first

PR #697, opened 2016-04-05
Add: Vertical lines to cards

andygout:
cc @ironsidevsquincy 

Will require changes in this `n-section` PR to hide unwanted card rules on far left of section: https://github.com/Financial-Times/n-section/pull/11

##### From:
from

##### To:
to

### DIFF
--- a/shared/components/_card.scss
+++ b/shared/components/_card.scss
@@ -17,6 +17,35 @@
 		background-color: oColorsGetColorFor(card, border);
 	}
 
+	// left rule
+	&:after {
+		content: '';
+		position: absolute;
+		width: 1px;
+		left: -5px;
+		top: 5px;
+		bottom: -5px;
+		background-color: oColorsGetColorFor(card, border);
+		@include oGridRespondTo(M) {
+			left: -7px;
+			bottom: -10px;
+		}
+
+		.fast-ft &,
+		.section--myft &,
+		.section--video & {
+			display: none;
+		}
+	}
+
+	&.card--picture-story:after {
+		display: none;
+	}
+
+	.section--opinion & {
+		@include oColorsFor(card, background);
+	}
+
 	.fast-ft & {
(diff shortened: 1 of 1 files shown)


shared/components/_card.scss
```diff
@@ -17,6 +17,31 @@
...
+		content: '';
+		position: absolute;
+		width: 1px;
+		left: -5px;
+		top: 5px;
+		bottom: -5px;
+		background-color: oColorsGetColorFor(card, border);
+		@include oGridRespondTo(M) {
+			left: -7px;
+			bottom: -10px;
```
ironsidevsquincy: instead of hard-coding values, it would be better to use some ratio of `oGridGutter`

shared/components/_card.scss
```diff
@@ -17,6 +17,32 @@
...
+		@include oGridRespondTo(M) {
+			bottom: oGridGutter(M) / -2;
+			left: oGridGutter(M) / -2;
+		}
+
+		.fast-ft &,
+		.section--myft &,
+		.section--video & {
+			display: none;
+		}
```
ironsidevsquincy: can combine these selectors with the one below, as they're doing the same thing (hiding the rule)

shared/components/_card.scss
```diff
@@ -17,6 +17,32 @@
...
+		position: absolute;
+		width: 1px;
+		top: oGridGutter() / 2;
+		bottom: oGridGutter() / -2;
+		left: oGridGutter() / -2;
+		background-color: oColorsGetColorFor(card, border);
+
+		@include oGridRespondTo(M) {
+			bottom: oGridGutter(M) / -2;
+			left: oGridGutter(M) / -2;
```
ironsidevsquincy: should there be a `top`?

ironsidevsquincy:
Bar comments, :+1:

PR #698, opened 2016-04-05
Remove: Card rule color and use case

andygout:
cc @ironsidevsquincy 

Card border use case color is now being set in `n-card` and its presence here will override the use case.

`n-card` PR which adds use case: https://github.com/Financial-Times/n-card/pull/44


ironsidevsquincy:
:+1:

PR #699, opened 2016-04-05
Different colour top rule for opinion cards

ironsidevsquincy:
Dependent on https://github.com/Financial-Times/n-card/pull/46


andygout:
👍

PR #701, opened 2016-04-06
Ad nicer names

adgad:

Show new responsive ad layout if adsNewProposition flag on, and map some nicer URL names to the targeting params used by ops (for the purpose of external demos)

### DIFF
--- a/server/routes/front-page.js
+++ b/server/routes/front-page.js
@@ -12,7 +12,7 @@ const getAdsLayout = (requestedLayout, flags) => {
 	//map some url params to existing ad layout names
 
 	if(flags.adsNewProposition && !requestedLayout) {
-		requestedLayout = 'new';
+		requestedLayout = 'responsive';
 	}
 	const mapping = {
 		'responsive': 'highimpact',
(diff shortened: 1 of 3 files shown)


server/routes/front-page.js
```diff
@@ -12,8 +12,18 @@ const contentMissing = data => {
...
-	return requestedLayout || 'default';
+const getAdsLayout = (requestedLayout, flags) => {
+	//map some url params to existing ad layout names
+
+	if(flags.adsNewProposition && !requestedLayout) {
+		requestedLayout = 'responsive';
+	}
+	const mapping = {
+		'responsive': 'highimpact',
+		'non-responsive-1': 'billboards',
```
VladDubrovskis: do we still need this? does not seem to be used anywhere else?

adgad: So this is a hack to allow us to use nicer URLs (e.g. `?adsLayout=responsive`) to send out externally (without having to change all the ad line items, which are still targeted to the less useful nlayout names like highimpact.

VladDubrovskis:
:+1:

PR #702, opened 2016-04-06
Implement: Compatibility with new n-section (v2.1.2)

andygout:
cc @ironsidevsquincy

### DIFF
--- a/client/components/myft/myft.js
+++ b/client/components/myft/myft.js
@@ -10,6 +10,9 @@ import sessionClient from 'next-session-client';
 import getSection from '../../../config/sections/index';
 import { user as query } from '../../../config/queries';
 import MyftPromo from '../../../shared/components/myft-promo/myft-promo';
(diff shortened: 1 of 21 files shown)


client/components/myft/myft.js
```diff
@@ -10,6 +10,9 @@ import sessionClient from 'next-session-client';
 import getSection from '../../../config/sections/index';
 import { user as query } from '../../../config/queries';
 import MyftPromo from '../../../shared/components/myft-promo/myft-promo';
+import Components from '@financial-times/n-section';
+
+Components.MyftPromo = MyftPromo;
```
ironsidevsquincy: not sure when this will be run, maybe when the js is evaluated or maybe when this module is imported. Either way, should move into on exported function below (and in the `if` statement)

ironsidevsquincy:
apart from comment, :+1:

PR #703, opened 2016-04-06
Hide card rule in first row of section

ironsidevsquincy:
Because of https://github.com/Financial-Times/n-section/pull/15


andygout:
👍

PR #704, opened 2016-04-06
Remove redundant border zeroing

keirog:
The bottom border has been removed from the default n-card, so we don't need this now. (Also this fixes the missing bottom border in the concept card.)


andygout:
👍

PR #706, opened 2016-04-06
Prevent: UglifyJsPlugin warnings from displaying

andygout:
cc @ironsidevsquincy @matthew-andrews 

Dependent on `n-makefile` PR: https://github.com/Financial-Times/n-makefile/pull/41


matthew-andrews:
could we just turn warnings off always?

matthew-andrews:
or is there a good reason to want them sometimes?

andygout:
My only thought for having the option of keeping them is if you wanted to work through the warnings and fix them, but I suppose you could always hard-code your local `webpack.config.js` so that `UglifyJsPlugin` is given no arguments and displays them.

matthew-andrews:
i don't really mind…  i think to begin with just suppressing them is fine and then we can always add options later to make it more configurable if people have a specific use case?  [the fewer options+switches we have to maintain the better]

andygout:
Sounds good; have amended `n-makefile` PR: https://github.com/Financial-Times/n-makefile/pull/41 and am closing this PR.

PR #707, opened 2016-04-06
Override the default page background color

keirog:
This is so we can use Sass' `mix` function in our components, resulting 
in solid background colors that are visually the same as RGBA-transparent 
colors, without relying on RGBA browser support.


ironsidevsquincy:
:+1:

PR #710, opened 2016-04-07
Update: Opinion section layout (frontpage audit)

andygout:
cc @ironsidevsquincy 

Stories 2-5 are now vertically balanced (and as a result stories 3 & 4 have switched position).

Default and S layouts remain the same.

##### From (L - XL):
from l-xl

##### To (L - XL):
to l-xl

##### From (M):
from m

##### To (M):
to m


ironsidevsquincy:
:+1:

PR #712, opened 2016-04-07
Remove: Duplicate imports from n-section

andygout:
cc @ironsidevsquincy

### DIFF
--- a/client/components/myft/myft.js
+++ b/client/components/myft/myft.js
@@ -3,7 +3,6 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 import { crossDomainFetch } from 'n-jsonp';
-import { Section, Content } from '@financial-times/n-section';
 import * as myFtUi from 'next-myft-ui';
 import sessionClient from 'next-session-client';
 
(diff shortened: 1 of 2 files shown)


client/components/myft/myft.js
```diff
@@ -68,7 +67,7 @@ const handleResponse = (myFtContainerEl, myftClient, flags, response) => {
 		section.layout = section.layout.map(changeComponent.bind(null, followed.length < 2));
 	}
 
-	ReactDOM.render(<Section {...section} />, myFtContainerEl);
+	ReactDOM.render(<Components.Section {...section} />, myFtContainerEl);
```
ironsidevsquincy: minor thing, but as `Components` is an object, should be lower case

server/routes/front-page.js
```diff
@@ -49,7 +47,7 @@ export default region => (req, res) => {
 
 	const renderParams = {
 		layout: 'wrapper',
-		Section,
+		Section: Components.Section,
```
ironsidevsquincy: ditto

ironsidevsquincy:
aside from comments, :+1:

PR #727, opened 2016-04-08
Fix: Vertical card rules for big story

andygout:
cc @ironsidevsquincy 

Prevents lines from spilling out from bottom of background.

##### From:
from

##### To:
to


ironsidevsquincy:
Will need to put this in the ie8 stylesheet too - https://github.com/Financial-Times/next-front-page/blob/master/shared/components/_card-ie8.scss

ironsidevsquincy:
:+1:

PR #729, opened 2016-04-08
Add: Card left rule to picture story cards

andygout:
cc @ironsidevsquincy 

##### From:
from

##### To:
to


ironsidevsquincy:
:+1:

PR #730, opened 2016-04-11
Use: Specific markets data use case for color

andygout:
cc @ironsidevsquincy 

Will use this specific `markets-data` use-case from `next-sass-setup`: https://github.com/Financial-Times/next-sass-setup/pull/127 given frontpage now overrides the use-case that it previously used: https://github.com/Financial-Times/next-front-page/blob/master/client/main.scss#L18

##### From (note colour of text for 'FTSE 100'):
from

##### To:
to


ironsidevsquincy:
:+1:

PR #733, opened 2016-04-12
Amend: GraphQL query limits

andygout:
cc @ironsidevsquincy 

Bring limits in line with what is actually required for frontpage.


ironsidevsquincy:
:+1:

PR #734, opened 2016-04-12
Revise: Highlights + Most Read sections L layout

andygout:
cc @ironsidevsquincy 

Requires review and approval from PO.

##### From:
from 1219

##### To (at 1219px):
to 1219

##### From (at 980px):
from 980

##### To (at 980px):
to 980


ironsidevsquincy:
:+1:

PR #736, opened 2016-04-14
Ad transitions and improvements

adgad:
/cc @ironsidevsquincy @VladDubrovskis @andrewgeorgiou1981 @leggsimon 
Various ad proposition improvements behind another flag.

* Transition the ad loading
* Add patterned background on top ad
* Full bleed responsive ad in middle
ad-loading


VladDubrovskis:
Not too convinced by the "vw" but 👍 as its behind flag and we can test it

PR #739, opened 2016-04-18
Add: o-card class to myFT promo box

andygout:
cc @ironsidevsquincy 

Required for when `o-card` is used on the frontpage. Has no effect by itself so can be safely merged now in anticipation.


ironsidevsquincy:
👍

PR #741, opened 2016-04-21
Remove `mixin-exists` check

keirog:
Sass won't let us declare mixins within `@if` anymore (technically we weren't ever supposed to do it, but now it errors).

(The mixin shouldn't already exist for this app anyway, so removing the check should be safe.)


andygout:
Pending `o-header` fix to pass build 👍

PR #742, opened 2016-04-22
Add: Back-to-top link for default layout

andygout:
cc @ironsidevsquincy 

back-to-top

### DIFF
--- a/client/components/back-to-top/_back-to-top.scss
+++ b/client/components/back-to-top/_back-to-top.scss
@@ -14,7 +14,7 @@
 			padding-right: 10px;
 		}
 
-		&:hover::before {
+		&:hover:before {
 			@include nextIcon(arrow-up, getColor('cold-1'), 10, $icon-only: true);
 		}
 	}
--- a/client/components/back-to-top/_back-to-top.scss
+++ b/client/components/back-to-top/_back-to-top.scss
@@ -3,8 +3,8 @@
 	text-align: center;
 
 	.back-to-top__link {
-		display: block;
-		padding: 15px 0;
+		display: inline-block;
+		padding: 15px;
 		border-bottom-color: transparent;
 
 		&:before {
--- a/client/components/back-to-top/_back-to-top.scss
+++ b/client/components/back-to-top/_back-to-top.scss
@@ -15,7 +15,7 @@
 		}
 
 		&:hover::before {
-			@include nextIcon(arrow-up, getColor('cold-1'), 10);
+			@include nextIcon(arrow-up, getColor('cold-1'), 10, $icon-only: true);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


client/components/back-to-top/_back-to-top.scss
```diff
@@ -0,0 +1,25 @@
...
+		border-bottom-color: transparent;
+
+		&:before {
+			@include nextIcon(arrow-up, getColor('teal-1'), 10);
+			content: '';
+			margin-left: -5px;
+			padding-right: 10px;
+		}
+
+		&:hover::before {
```
ironsidevsquincy: single colon

client/components/back-to-top/_back-to-top.scss
```diff
@@ -0,0 +1,25 @@
...
+			@include nextIcon(arrow-up, getColor('cold-1'), 10);
```
ironsidevsquincy: `nextIcon` has a fourth argument, `$icon-only`, which will just output the icon and exclude the width and height (as you've already set them above)

https://github.com/Financial-Times/next-sass-setup/blob/64a9065f518bc726ead3c46b17fbcb4b59a1d687/_icons.scss#L1

client/components/back-to-top/_back-to-top.scss
```diff
@@ -0,0 +1,25 @@
+.back-to-top {
+	padding: 15px 10px 10px;
+	text-align: center;
+
+	.back-to-top__link {
+		display: block;
```
ironsidevsquincy: wonder if we want to make this full width? might it interfere with people scrolling down the screen?

andygout: Given it's at the bottom of the page does not feel so much of a problem (as opposed to having this link dotted throughout the page), and devices discriminate between the scroll/slide action of a finger and a press action, so on that basis feel keeping the touch target area big is the better option.

client/components/back-to-top/_back-to-top.scss
andygout: Is it okay to use CSS3 syntax if I don't display it on IE8? https://css-tricks.com/almanac/selectors/a/after-and-before/

ironsidevsquincy: it's a touchy subject, but I'd stick to single quotes for everything

views/partials/back-to-top.html
```diff
@@ -0,0 +1,3 @@
+<div class='back-to-top'>
+	<a href='#' class='back-to-top__link' data-trackable='back-to-top'>Back to the top</a>
```
ironsidevsquincy: don't know the general policy on using just `#`, but feels a bit wrong. Better link actually to something. The skip links at the top use `#primary-nav`, could use that (though looks like that's not actually on the page, so currently broken)

andygout: Will fix `#primary-nav` in `o-header` and then add that as the 'back-to-top' link.

ironsidevsquincy:
👍

PR #745, opened 2016-04-22
Stop sending flags to graphql

wheresrhys:
It doesn't appear to be used for anything @ironsidevsquincy


ironsidevsquincy:
It's used to send the `mockFrontPage` flag, which returns mocked out data

PR #746, opened 2016-04-22
passing un-headerified mock mockFrontPage flag to graphql

wheresrhys:
@ironsidevsquincy

### DIFF
--- a/server/libs/graphql-poller.js
+++ b/server/libs/graphql-poller.js
@@ -24,15 +24,17 @@ let queryResults = {
 let readyPromise;
 
 const pollData = (query, variables, name, flags = {}) => {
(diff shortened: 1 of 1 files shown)


ironsidevsquincy:
Great, thanks 👍

server/libs/graphql-poller.js
```diff
@@ -24,15 +24,17 @@ let queryResults = {
...
 			method: 'POST',
-			body: JSON.stringify({ query, variables }),
+			body: JSON.stringify({
+				query,
+				variables
+			}),
 			headers: {
 				'Content-Type': 'application/json',
-				'X-Flags': xFlagsHeader
+				'FT-Mock-Front-Page': flags.mockFrontPage
```
ironsidevsquincy: Can it be just `FT-Is-Mock`, or something (it's not necessarily specific to the front page)

wheresrhys: good call

PR #749, opened 2016-04-25
Added code to ie8 CSS to display the sign-in link in the header

adambraimbridge:
cc @laurieboyes Thanks

image


laurieboyes:
yay 👍

PR #759, opened 2016-04-29
Set: o-typography resolution (^3.2.0)

andygout:
cc @ironsidevsquincy


andygout:
`o-typography` now resolved so this PR no longer required.

PR #763, opened 2016-05-03
Fix: HTML background color

andygout:
cc @ironsidevsquincy @keirog 

Overriding the default page background color must come after `n-section` is imported, else `n-section` imports `n-card`, which then imports `next-sass-setup`, which will supersede the page background color use case (in `_colors-use-cases.scss`).


ironsidevsquincy:
Yeah, this is tricky. Can't put it before or else the myft concept card's will be the wrong colour. So until we have a unified bg colour across the site, hard-coded it in https://github.com/Financial-Times/next-front-page/pull/762

keirog:
Sorry yep, we needed that. I'll catch up with you shortly about possible workarounds

PR #767, opened 2016-05-04
Remove full bleed and use transition to collapse

adgad:
@VladDubrovskis @andrewgeorgiou1981 @leggsimon @aintgoin2goa @andygout 
* full bleed mid-ad was a bit buggy so remove for now
* when ad collapses, transition rather than jump


VladDubrovskis:
👍

PR #787, opened 2016-05-10
Rhys/n ui2

wheresrhys:
Adds 1000k to main.css unminified/gzipped (I think I'm accidentally including grid twice or something), but acceptable for now I think as doesn't affect start render. Will sort that out when I move n-layout & sass-setup in to here... tomorrow :smile:

@ironsidevsquincy

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -1,36 +1,22 @@
-import oDate from 'o-date';
-import layout from 'n-layout';
-import setup from 'next-js-setup';
 import nVideo from 'n-video';
-import prompts from 'n-message-prompts';
-import myFtClient from 'next-myft-client';
-import * as myFtUi from 'next-myft-ui';
-
+import {bootstrap} from 'n-ui';
 import highlightDomPath from './components/highlight-dom-path/highlight-dom-path';
 import scrollDepth from './components/scroll-depth/scroll-depth';
 import marketsData from './components/markets-data/markets-data';
 import headerTabs from './components/header-tabs/header-tabs';
 import myft from './components/myft/myft';
 
-setup.bootstrap(({ flags }) => {
-	// NOTE: make sure we init myft client *before* n-layout
-	const clientOpts = [];
-	flags.get('follow') && clientOpts.push({relationship: 'followed', type: 'concept'});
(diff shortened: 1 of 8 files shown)


client/main.scss
```diff
@@ -1,8 +1,7 @@
-$next-sass-setup-is-silent: true;
-@import 'next-sass-setup/main';
+$n-ui-is-silent: true;
 
-@import 'n-layout/non-critical';
-@import "n-message-prompts/main";
+@import 'n-ui/non-critical';
```
ironsidevsquincy: `n-ui/critical-layout` and `n-ui/non-critical-layout`? to make it clear what this does

client/main.js
```diff
@@ -1,36 +1,22 @@
...
 		placeholder: true,
 		classes: ['video'],
 		selector: '.js-video'
 	});
-	oDate.init();
-	prompts.init();
+
 	highlightDomPath();
 	scrollDepth(flags);
-	myFtUi.init({ anonymous: !(/FTSession=/.test(document.cookie)) });
```
ironsidevsquincy: where does this get init-ed now? and previously we had to init the client before this

client/main.scss
wheresrhys: the connection to layout is gonna be short-lived as I will get rid of n-layout and bring all its code in here, so these files will be the critical and non-critical bits of n-ui itself

client/main.js
wheresrhys: https://github.com/Financial-Times/n-ui/blob/master/main.js

As I roll out to all user-facing apps I may make initing myft an opt in rather than on by default

ironsidevsquincy:
Looks good, but if we could hold off a bit, just while @i-like-robots finalises his plugin (https://github.com/Financial-Times/n-makefile/pull/56), then I think that'll fix the duplication issue

wheresrhys:
OK I'll leave this app til last, but don't want to be blocked for too long. I feel like I've been in rebase purgatory recently.

ironsidevsquincy:
Ok, we now have a shiny plugin to create multiple css files from a single entry point (https://github.com/i-like-robots/css-extract-block-webpack-plugin), so go ahead and merge this, and I'll add the plugin after to fix the duplication

PR #788, opened 2016-05-10
Wrap myFT component in it's own flag

phamann:
Currently the `getViewed()` request that the myFT component makes is failing. This is causing issues throughout the stack. Whilst we diagnose the issue further we want to remove the pressure on the DB my disabling the component.


ironsidevsquincy:
👍

PR #813, opened 2016-05-19
dont show right image on opinion card

aintgoin2goa:


ironsidevsquincy:
if an image goes from, say `right` at one breakpoint to `bottom` later, wont this hide the image everywhere?

aintgoin2goa:
Yes - is there a better way?  I'm just guessing...

ironsidevsquincy:
Not ideal, but output css for each breakpoint, if it's `right`, display none, if it's `left`/`top`/`bottom`, display block

https://github.com/Financial-Times/n-card/blob/master/components/image/_image.scss

client/_main-shared.scss
```diff
@@ -90,4 +90,17 @@ $n-ui-is-silent: true;
...
+	// Make sure we don't show right-aligned images for opinion cards
+	.card--opinion[data-image="right"] .card__image-link {
+		display: none;
+	}
+
+	@each $layout-name in $_o-grid-layout-names {
+		@include oGridRespondTo($layout-name) {
+			.card--opinion[data-image~="#{$layout-name}--right"] .card__image-link {
+				display: none;
+			}
```
ironsidevsquincy: but will need to display again for `left`/`top`/`bottom`

PR #814, opened 2016-05-23
Update Makefile

adgad:
/cc @wheresrhys Am I right in saying these shouldn't be specified per app anymore?


wheresrhys:
ah yes :+1:

PR #821, opened 2016-05-25
Brexit Live Coverage and Build-up components

sumeetadur:
2 Brexit components behind flags. The plan for this PR is to turn on the flag for this build-up component in its first iteration:

screen shot 2016-05-26 at 12 14 44


config/pages.js
```diff
@@ -3,7 +3,7 @@ import sectionData from '../server/libs/section-data';
 
 const getPages = flags => {
 
-	return {
+	let pages = {
```
ironsidevsquincy: can't this be a `const`?

client/components/brexit-coverage/_brexit-coverage.scss
```diff
@@ -0,0 +1,80 @@
...
+
+		&:after {
+			content: '';
+			display: block;
+			clear: both;
+		}
+	}
+
+	.section-meta__title {
+		font-family: oFontsGetFontFamilyWithFallbacks(MetricWeb);
```
ironsidevsquincy: could use `oTypographySansDataBold(m, $load-progressively: true)`. You will need to override the sizes, as 16/22 isn't one of them, but it gives you progressively loaded fonts

client/components/brexit-buildup/_brexit-buildup.scss
```diff
@@ -0,0 +1,75 @@
...
+	.section-meta {
+
+		@mixin adjustHeader($logo-size, $font-size) {
+			$horizontal-spacing: ($logo-size / 6);
+
+			line-height: $logo-size;
+
+			.section-meta__title {
+				font-size: $font-size;
+				@media (min-width: 360px) {
```
ironsidevsquincy: what is this breakpoint?

client/components/brexit-buildup/_brexit-buildup.scss
```diff
@@ -0,0 +1,75 @@
...
+						content: '';
+						position: absolute;
+						background-size: contain;
+						background-repeat: no-repeat;
+						background-position-x: right;
+						height: $logo-size;
+						width: $logo-size;
+						top: -2px;
+						left: -($logo-size + $horizontal-spacing - 10);
+						background-image: url('https://image.webservices.ft.com/v1/images/raw/http%3A%2F%2Fig.ft.com%2F_other%2Fbrexit-stream%2Flogo.png?source=next-stream-page&width=#{$image-size}&height=#{$image-size}&format=png');
```
ironsidevsquincy: use the geebee domain for images (is H2 enabled), e.g.

`https://next-geebee.ft.com/image/v1/images...`

client/components/brexit-buildup/_brexit-buildup.scss
```diff
@@ -0,0 +1,75 @@
...
+		@include oGridRespondTo(S) {
+			@include adjustHeader(40px, 20px);
+		}
+	}
+
+	.section-meta__follow {
+		display: none;
+		top: 0;
+
+		@media (min-width: 300px) {
```
ironsidevsquincy: another odd breakpoint?

client/components/brexit-coverage/_brexit-coverage.scss
```diff
@@ -0,0 +1,79 @@
...
+		&:before {
+			display: none;
+		}
+	}
+
+	.section-meta__follow {
+		display: none;
+		top: 0;
+
+		@media (min-width: 300px) {
```
ironsidevsquincy: bp?

ironsidevsquincy:
Apart form comments, 👍

client/components/brexit-buildup/_brexit-buildup.scss
sumeetadur: These breakpoints are to hide the Brexit logo and follow button on small devices as they were showing on multiple lines with the dotted lines between them

ironsidevsquincy: I wonder if there's anything we can do tho handle that situation; smaller text, better formatting when on two lines

It's just quite a slippery slope if we start adding intermediate breakpoints, ideally we'd only ever use the ones we've defined (which isn't to say we can't if we really need to)

sumeetadur: I reduced the font size to make it work better but I've been wondering if we could define an 'XS' breakpoint in o-grid at or around 360px, I've thought about it in a previous project but thought it might have been a one-off, but it's definitely useful in this case.

ironsidevsquincy: I think at one point we did actually have a XS, at 320 I think. Not against that, but it's definitely a Next-wide decision (we previously had a XXL breakpoint just for the front page, which caused issues with its inconsistency)

Worth talking through with UX/D about

config/layouts/brexit-buildup.js
```diff
@@ -0,0 +1,76 @@
...
+		type: 'Row',
+		components: [
+			{
+				type: 'Column',
+				colspan: { default: 12, M: 4 },
+				components: [
+					{
+						type: 'Content',
+						size: 'large',
+						image: createImageConfig({ default: 12 }, { default: 'top' }, [196, 276, 365, 660])
```
ironsidevsquincy: As the column decreases to 4 grid units at `M`, you want to pass that on to the `createImageConfig` function

```createImageConfig({ default: 12, M: 4 }, { default: 'top'...`

PR #827, opened 2016-05-31
bower resolutions

ifyio:


aintgoin2goa:
Is this still needed?  I'm guessing probably not, so I'm gonna close it unless I hear otherwise...

PR #836, opened 2016-06-03
Update Brexit poll countdown text

sumeetadur:
screen shot 2016-06-03 at 14 58 54


sumeetadur:
@ironsidevsquincy could you have a look?

ironsidevsquincy:
Do we know how many days before we're going to show the countdown? i.e. could we add, say

```
if (daystogo > 10) {
    return 'Latest polls';
} else if (daystogo > 1) {
    return `${daystogo} days until the referendum`;
}
```

To the existing function, rather than doing a code change

sumeetadur:
Unfortunately we don't know, it's an editorial decision that hasn't been made yet

ironsidevsquincy:
Shame, 👍

PR #840, opened 2016-06-05
De-dupe Brexit components

sumeetadur:


sumeetadur:
@ironsidevsquincy could you have a look?

matthew-andrews:
Going to ship this in the hope it makes things better……

PR #841, opened 2016-06-06
Delete brexit components

matthew-andrews:


andygout:
👍  Presuming Brexit team will need to reinstate this down the line.

PR #842, opened 2016-06-06
BrexitPrimaryResults component should render nothing when no data

callumlocke:
Addresses https://app.getsentry.com/nextftcom/ft-next-front-page/issues/132254895/


commuterjoy:
@sumeetadur @callumlocke can we get the incident report in before this is merged please?

https://github.com/Financial-Times/next-bugs

callumlocke:
As far as I'm aware, the incident this morning was nothing to do with this code. This is a separate, in-development feature behind a flag which was never supposed to be turned on except in development.

callumlocke:
correction: this code DID contribute to downtime, in that it errored and the site went down! but the point is this code was never meant to be called at all, the real bug is elsewhere. This PR is just a defensive move against the same thing happening again.

PR #843, opened 2016-06-06
Make pages config immutable

callumlocke:


ironsidevsquincy:
How big is the `immutable` library, i.e. how much does it add to our JS bundle?

matthew-andrews:
isn't this server side only or have i misunderstood? @ironsidevsquincy

ironsidevsquincy:
Think config will be bundled up with front end code

callumlocke:
yeah this is server-side only, it's in the config

ironsidevsquincy:
https://github.com/Financial-Times/next-front-page/blob/master/client/components/myft/myft.js#L10

callumlocke:
well well. my mistake

ironsidevsquincy:
Hang on, maybe not. Might have confused everything. Have a look in the build `main.js` to make sure

callumlocke:
No you're right, it's in the client bundle via [this line](https://github.com/Financial-Times/next-front-page/blob/f187697ed18b62badbb3c8682f2cfffb8ed1d859/client/main.js#L7).

Immutable.js 3.8.1 comes to about [16.1KB minified and gzipped](https://cdnjs.cloudflare.com/ajax/libs/immutable/3.8.1/immutable.min.js).

We could alternatively forget Immutable.js and just revert `getPages` so it's a pure function again, making sure it never exposes a reference to the original array. (i.e. it would return a `.slice()`.)

I guess the questions are: how do you feel about the practice of formalising/enforcing immutability using a library (as opposed to just relying on careful coding to avoid this class of bug), and do you think it's worth adding 16.1KB to the clientside bundle? (Also, would it be just for this one case, or do you think we'd end up using it in more places?)

ironsidevsquincy:
I'm actually all for immutable, mutability has caused me (and I'm sure others) a whole heap of hard-to-track-down issues in the past. However, I think we might need a wider team conversation before switching over. So I'd say, for now, switch it to a pure function and we'll put it on our dev discussion list

callumlocke:
ok this should be good now, and I've PR'd an incident report explaining everything in detail. I'd like to get this merged tomorrow morning so we can try reintroducing the Brexit code.

ironsidevsquincy:
👍

bjfletcher:
We used `Object.freeze` to a great effect in a previous project. Just a couple of things:

* it freezes just the first-level properties;
* it's not included in the `polyfill.io` list (yet?)

bjfletcher:
https://github.com/Financial-Times/polyfill-service/issues/232

@JakeChampion: I think `Object.freeze = indentity` is a good proposal, it is actually something I do within the `Symbol` polyfill and forgot to pull out into it's own polyfill - https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Symbol/polyfill.js#L22

PR #851, opened 2016-06-07
Restore Brexit components

callumlocke:
@matthew-andrews, @ironsidevsquincy: I think this is safe, please take a look


matthew-andrews:
looks nice!

callumlocke:
OK I'm going to merge and keep an eye on it

PR #860, opened 2016-06-09
update flag name to allow pre-roll ads, and update o-video version

VladDubrovskis:
/cc @adgad @leggsimon @andrewgeorgiou1981 @lc512k


andrewgeorgiou1981:
👍

PR #864, opened 2016-06-10
Adapt: Back-to-top link to work with new header

andygout:
cc @i-like-robots  

The new header does not have an obvious `id` for the link to latch onto, so have gone for a 'one size fits all' approach. Only difference will be on the control where the link goes to the very top of the page, displaying Markets Data strip, rather than just below it.


i-like-robots:
Is this behaviour consistent between browsers? Also there are hidden skip links that also reference this ID... so it has to exist for them too.

PR #868, opened 2016-06-14
Add root:true to eslint config

callumlocke:
By default eslint walks right up the filesystem and merges all the configs it can find. The `"root": true` setting tells it to stop here.

reason: it's common to have a personal global config at `~/.eslintrc` for your own stuff. But when you're working in a project that defines its own config, you want the linter to respect that project's config, not merge it with your personal one.

http://eslint.org/docs/user-guide/configuring


aintgoin2goa:
👍 Feel free to merge this if you still need it

PR #873, opened 2016-06-16
Hack to unhide BrexitSecondary lead image

callumlocke:
@kavanagh this hack solves the immediate problem


kavanagh:
Ok, I'm happy if @andygout is happy. We can revert later if we find a a better solution.

andygout:
Hack is the word.

👍

ironsidevsquincy:
Ugh, why is that in the CSS in the first place? Images can be hidden in the layout configuration

PR #875, opened 2016-06-16
Hack to unhide BrexitSecondary lead image (#873)

wheresrhys:
@ironsidevsquincy 

I messed up rebasing - webpack is the only meaningful change here.. honest

Does it even matter if I got the main.core.js bit wrong?


ironsidevsquincy:
For now, probably not, it'll just have empty space

PR #881, opened 2016-06-20
BrexitLive component

callumlocke:
This sets up the `BrexitLive` component (formerly named `BrexitPrimaryInteractive`), an interactive live data widget that goes inside the brexit-primary section (the black panel to be shown on referendum night). It's been rewritten so all the clientside code is now vanilla JS with granular DOM updates.

There is still more work to do on this before it goes live, but it would be good if @ironsidevsquincy and/or @andygout could have a look today and flag up anything controversial in the code.

image


andygout:
Nothing untoward from my perspective, but def worth getting @ironsidevsquincy 's eagle eye over this.

shared/components/brexit-live/_brexit-live.scss
```diff
@@ -0,0 +1,212 @@
...
+			position: absolute;
+			text-transform: uppercase;
+			white-space: nowrap;
+
+			> span {
+				font-weight: 400;
+			}
+		}
+
+		&__notes {
```
ironsidevsquincy: I do like this pattern, but it makes it very hard to grep

Also, do things need to be nested? i.e. isn't `.brexit-live__result__notes` specific enough?

ironsidevsquincy:
👍 , though haven't followed the JS in detail...

shared/components/brexit-live/_brexit-live.scss
callumlocke: Sass renders it to `.brexit-live__result__notes`. There's no nesting. I think.

ironsidevsquincy: Ah, ok

callumlocke:
OK. I'm going to merge this then so other people can test it with the flag on

PR #888, opened 2016-06-21
Brexit primary style tweaks

callumlocke:


client/components/brexit-primary/_brexit-primary.scss
```diff
@@ -1,66 +1,101 @@
 .section--brexit-primary {
-	position: relative;
-	padding-bottom: 20px;
-	margin-top: 20px;
+	$brexit-pale-grey: getColor('grey-tint1');
+	$brexit-buff: #ffc99f;
```
andygout: Am guessing no color in the palette for this?

andygout:
👍

client/components/brexit-primary/_brexit-primary.scss
callumlocke: Correct, although there's a chance it might be added in future if this black panel thing gets promoted into a permanent section type

andygout: Makes sense.

callumlocke: there is also an ongoing controversy in the newsroom over whether this colour can be called 'gold' or not :) personally I think it's [buff](https://en.wikipedia.org/wiki/Buff_(colour))

PR #889, opened 2016-06-22
Disable lazy-load for the big image in BrexitPrimary

callumlocke:
This just disables it for the main image. (We still want to lazy-load headshots.)


ironsidevsquincy:
👍

PR #892, opened 2016-06-22
More Brexit CSS tweaks

callumlocke:


callumlocke:
@andygout @ironsidevsquincy – if there are further minor Brexit-related changes like this over the next 72 hours (likely), can we peer-review them ourselves within IG (me and @kavanagh) instead of hassling you guys?

By minor changes, I mean: changes that only touch a Brexit-related file, and which don't add any deps, and which we're confident about and pass CI obviously.

ironsidevsquincy:
Sure thing

PR #893, opened 2016-06-22
Switch international edition to get top stories from list

ironsidevsquincy:
Rename query to `topStory` and `topStories`

@andygout


andygout:
Ah, so much simpler!

👍 once build passing.

PR #896, opened 2016-06-23
Finishing brexit live component

callumlocke:


andygout:
👍  Pending passing tests.

Lots of commented out code in `shared/components/brexit-live/_brexit-live.scss` maybe worth deleting unless you want to keep for some reason?

PR #902, opened 2016-06-23
Don't poll when tab is idle

callumlocke:


client/components/brexit-live/brexit-live.js
```diff
@@ -193,6 +193,13 @@ export default (selector = '.brexit-live') => {
 			const updateData = () => {
 				if (destroyed) return;
 
+				// skip this poll if tab is hidden
+				if (document.hidden) {
+					console.log('tab idle; skipping poll...');
```
commuterjoy: naughty `console.log`

commuterjoy:
+1 apart from my comment

client/components/brexit-live/brexit-live.js
callumlocke: will do ta

PR #903, opened 2016-06-23
fix result display when count is zero

callumlocke:
also bundled in this PR: added some o-tracking


client/components/brexit-live/brexit-live.js
```diff
@@ -283,23 +283,43 @@ export default (selector = '.brexit-live') => {
...
+				event.stopPropagation();
+
 				pause();
+
 				store.dispatch({
 					type: SELECT_TAB,
 					index: Number(target.getAttribute('data-tab-index')),
 				});
+
+				document.body.dispatchEvent(new CustomEvent('oTracking.event', {
```
commuterjoy: i would put a `data-trackable="national"`and `data-trackable="markets"` in your HTML instead

client/components/brexit-live/brexit-live.js
```diff
@@ -283,23 +283,43 @@ export default (selector = '.brexit-live') => {
...
+				event.stopPropagation();
 			});
 
 			// when user clicks progress control, toggle play/pause
 			const onTogglePlayPause = () => {
 				if (store.getState().paused) play();
 				else pause();
+
+				document.body.dispatchEvent(new CustomEvent('oTracking.event', {
+					detail: {
```
commuterjoy: you could wrap an <button> or <a> around the SVG and do the same thing

client/components/brexit-live/brexit-live.js
callumlocke: issue is I'm stopping propagation and handling the event manually, so I figured that wouldn't work?

commuterjoy: Our click tracking happens on capture to avoid that problem

https://github.com/Financial-Times/n-instrumentation/blob/master/src/analytics/cta.js#L46

client/components/brexit-live/brexit-live.js
callumlocke: I can't see any mention of data-trackable on https://github.com/Financial-Times/o-tracking
where's it documented?

commuterjoy: http://financial-times.github.io/next/docs/developer-guide/naming-conventions/ 

it's just a next specific thing

callumlocke: oh I think I remember how it works now

commuterjoy:
+1

PR #904, opened 2016-06-24
Remove card rule work

ironsidevsquincy:
It's in [n-card now](https://github.com/Financial-Times/n-card/blob/master/components/_card.scss#L26-L51)


callumlocke:
+1

PR #907, opened 2016-06-25
Remove code no longer needed by BrexitLive component

callumlocke:
The BrexitLive component is now showing only the Markets data tab. The other tabs have been switched off and won't come back.

This PR removes some of the code no longer in use, including a constant animation loop that is now doing nothing.

image

NB. there will be a bigger PR after the weekend to remove all trace of the various Brexit components, once the flags have all been turned off. This is just a quick one for low hanging perf fruit


phamann:
👍 LGTM

PR #912, opened 2016-06-29
preconnect more of the things

wheresrhys:
@ironsidevsquincy


ironsidevsquincy:
Did have this at one point, but seeing as they're both JS loaded, quite late (and video is at the bottom), thought it would be better to not use a couple of resources on them

wheresrhys:
fair enough

PR #914, opened 2016-06-29
[DO NOT MERGE] Remove brexit components

sumeetadur:


.gitignore
```diff
@@ -5,6 +5,8 @@
 
 /public/__about.json
 /public/asset-hashes.json
+/public/head-ie8.css
+/public/head-ie8.css.map
```
ironsidevsquincy: no need for these (don't make head files for IE8)

sumeetadur: Thanks, just realised that myself. I had an old version of the `public` directory with those files in it.

callumlocke:
Forgot about this. Is it now irrelevant? Has this stuff been removed/fixed already?

PR #919, opened 2016-07-06
fix big story layout

aintgoin2goa:


ironsidevsquincy:
Should go https://github.com/Financial-Times/n-section/blob/master/components/column/_column.scss#L41

guessing that should be `background-color`, not `color`?

aintgoin2goa:
Aha!  Ok, thanks will move it

PR #923, opened 2016-07-06
Adds the regional news section

GlynnPhillips:
screen shot 2016-07-06 at 17 06 46

This is being added to allow for two AB tests.

1) If the reader is reading the UK edition display a UK news section
2) If the reader is reading the International edition and in the US
display a US news section and nothing if outside of the US

The `ukRegionalNews` or `usRegionalNews` flags can be toggled to see the
results. Although as toggler forces a flag to be on the geolocation aspect cannot be demo'ed using this so it doesn't matter which flag you turn on you will see UK news on the UK edition and US news on the international edition.

/cc @ironsidevsquincy @andygout @tom-parker


config/layouts/regional-top-stories.js
```diff
@@ -0,0 +1,84 @@
...
+		components: [
+			{
+				type: 'Column',
+				colspan: { default: 12, S: 4 },
+				components: [
+					{
+						type: 'Content',
+						size: 'medium',
+						isPictureStory: true,
+						image: createImageConfig({ default: 12 }, { default: 'top'}, [430, 210, 365]),
```
ironsidevsquincy: bit complicated this `createImageConfig`, but the first argument is how many cols with the image is a different breakpoints. seeing as the column itself is 4 cols wide at `S`, I'm guessing you want to do the same here

config/queries.js
```diff
@@ -125,6 +125,14 @@ const frontPage = `
 				... LiveBlogInfo
 			}
 		}
+		regionalTopStories: page(uuid: $regionSectionUuid) {
+			items(limit: 20) {
+				... Basic
+				... Extended
+				... Related
```
ironsidevsquincy: Do you need to get related content too?

GlynnPhillips: Good catch, forgot to change this

config/layouts/regional-top-stories.js
GlynnPhillips: Thanks thats good to know, I thought it was how relative to the parent col rather than the same. Will update this now.

ironsidevsquincy: Yeah, that would be better, but then it would have to know how big its parent is relative to its parent, which would also need to know big it is relative to its parent, etc. all the way up.

tom-parker:
Looks good to me

PR #924, opened 2016-07-07
Fixes the missing title on regional news

GlynnPhillips:
I left some code in an old flag which meant that the regional news
section was missing the heading.


config/sections/regional-top-stories.js
```diff
@@ -1,12 +1,10 @@
...
-	if(flags.regionalNews) {
-		if(region === 'UK') {
-			title = 'UK news';
-		} else {
-			title = 'US news';
-		}
+	if(region === 'UK') {
+		title = 'UK news';
+	} else {
+		title = 'US news';
```
ironsidevsquincy: wonder if it's better to switch these around, i.e. have UK the fallback. if for whatever reason region isn't available, you'd probably want it to show UK news, rather than US

GlynnPhillips: It seems that our default is the International edition which is why I replicated this logic.

https://github.com/Financial-Times/next-front-page/blob/master/server/init.js#L64

debugwand: you could use a const and ternary now you've dropped the wrapping if statement

`const title = (region === 'UK') ? 'UK news' : 'US news';`

GlynnPhillips: Good point, will do that now.

PR #925, opened 2016-07-07
Resolve bower conflict with o-typography

i-like-robots:
=]


lc512k:
@i-like-robots sorry to jump on this one, I needed your o-typography fix to rebuild the front page (so it would pick up some rather urgent mouseflow changes released today in n-ui)

PR #930, opened 2016-07-11
Fix Brexit chart heading issue on XL screens

sumeetadur:


shared/components/brexit-secondary-live/_brexit-secondary-live.scss
```diff
@@ -2,31 +2,50 @@
...
+				z-index: 1;
+			}
+		}
+
+		.brexit-secondary-live__tab {
+			background-color: oColorsGetPaletteColor('warm-5');
+			float: left;
+			font-size: 14px;
+			margin-top: -19px;
+			margin-left: calc(50% - 34px);
```
ironsidevsquincy: How does this look in IE8?

ironsidevsquincy:
Apart from query, 👍

PR #931, opened 2016-07-12
update the oVideo version to v1

VladDubrovskis:
/cc @lc512k


ironsidevsquincy:
Where are we with switching over to o-video?

VladDubrovskis:
There are still some mobile browser issues needs fixing, so once that is done. Also, reverting this one as seem to be breaking few things.

PR #936, opened 2016-07-13
Remove duplicates from regional news

GlynnPhillips:
Remove any duplicates from the regional news section that are also in
the brexit, top stories or opinion sections.

/cc @lc512k @tom-parker @adambraimbridge


tom-parker:
👍

PR #939, opened 2016-07-13
Record the number of top stories after dedup

GlynnPhillips:


server/libs/section-data.js
```diff
@@ -79,7 +81,8 @@ const getRegionalStories = (data, flags) => {
 	regionalStories = regionalStories.length < numRegionalStories ?
 		fillGapsWithPopular(regionalStories, alreadyUsedIds, data) :
 		regionalStories;
-
+	
+	metrics.histogram(`regionalTopStories.numberOfStories`, regionalStories.length);
```
commuterjoy: conventionally we use hyphens - i.e. `regional-top-stories.number-of-stories` - as it's easier to read in graphite

commuterjoy:
+1 apart from the hyphens thing

PR #941, opened 2016-07-14
Brexit markets link

sumeetadur:
Make the Brexit markets heading a link to the markets data page


client/components/brexit-secondary-live/brexit-secondary-live.js
```diff
@@ -96,7 +96,7 @@ export default (selector = '.brexit-secondary-live') => {
 
 					panelDiv.className = 'brexit-secondary-live__tab-panel';
 
-					li.innerHTML = `<span>${tabData.title}</span>`;
+					li.innerHTML = `<a href="http://markets.ft.com/data">${tabData.title}</a>`;
```
ironsidevsquincy: probably want a `data-trackable` attribute on this link

shared/components/brexit-secondary-live/_brexit-secondary-live.scss
```diff
@@ -25,6 +25,11 @@
 			text-transform: uppercase;
 			width: 68px;
 			z-index: 1;
+
+			a {
```
ironsidevsquincy: can we use a class instead of an element?

ironsidevsquincy:
👍

PR #942, opened 2016-07-14
Use consistent styles for Brexit heading links

sumeetadur:


shared/components/brexit-secondary-live/_brexit-secondary-live.scss
```diff
@@ -28,9 +28,13 @@
...
 		}
 
-		.brexit-secondary-live__charts_heading {
+		.brexit-secondary-live__charts-heading {
 			border-bottom: 0;
 			color: oColorsGetPaletteColor('grey-tint5');
+
+			&:hover {
+				color: oColorsGetPaletteColor('grey-tint4');
+			}
```
ironsidevsquincy: As this css is the same as the one above, could just abstract it into a single class and add that to elements that need the styling

ironsidevsquincy:
👍

PR #946, opened 2016-07-18
Adding the ability to move mid page advert

GlynnPhillips:
Moves the mid page advert above opinion


config/pages.js
```diff
@@ -1,25 +1,63 @@
...
+	if(flags.ukRegionalNews) {
+		regionalLayout = flags.ukRegionalNews;
+	} else if(flags.usRegionalNews) {
+		regionalLayout = flags.usRegionalNews;
+	} else {
+		regionalLayout = 'control';	
+	}
+	
+	// Get a CLONE of the layout array (we're about to mutate the cloned copy)	
+	const layout = pageLayout[regionalLayout.toLowerCase()];
```
aintgoin2goa: I think you need to clone this otherwise any changes you make below will be applied to the source object

PR #949, opened 2016-07-20
Swap popular from hui to most recent stories

GlynnPhillips:
When the regional news section needs to fill gaps due to de-duplication
use the most recent stories from a region instead of the HUI popular
articles


config/queries.js
```diff
@@ -132,7 +132,7 @@ const frontPage = `
 				... OpinionData
 			}
 		}
-		popularFromHui(country: $huiUrl, limit: 10, period: "last-1-day") {
+		regionalRecentStories: search(termName: "idV1", termValue: $regionSectionIdV1) {
```
ironsidevsquincy: Is that enough, or does it need to be `metadata.idV1`? Just basing it on https://github.com/Financial-Times/next-graphql-api/blob/master/server/lib/backend-adapters/fast-ft.js#L20

GlynnPhillips: It seems to be working 😕 But I will try with `metadata.idV1` as well and see if theres any difference.

GlynnPhillips: It appeared to work both ways, so I have switched it to `metadata.idV1` to keep it in line with other instances

PR #952, opened 2016-07-21
Adds additional regions to the regional news

GlynnPhillips:
Below is some pseudocode explaining what should appear where but the
logic for knowing what edition and location is handled by the flagsi /
AB test.

IF user is on UK edition and in the UK
	OUTPUT UK regional news section

IF user is on international edition and in the UK
	OUTPUT European regional news section

IF user is on international edition and in the Europe
	OUTPUT European regional news section

IF user is on international edition and in the UK
	OUTPUT US regional news section

IF user is on international edition and in the Asia Pacific
	OUTPUT Asian regional news section

IF user is on international edition and not from UK, US, Europe or Asia
Pacific
	OUTPUT No regional news section

IF user is on UK edition and not from UK
Pacific
	OUTPUT No regional news section


ironsidevsquincy:
Could we encapsulate those rules as tests?

GlynnPhillips:
Yeah I was thinking of looking at that this afternoon but just need to work out how much value they will add as main logic is handled by the flags.

GlynnPhillips:
@ironsidevsquincy I have added these tests which cover the logic that is actually happening within the front-page app.

https://github.com/Financial-Times/next-front-page/pull/952/files#diff-a1a1735689a5dcf4971523973c44b1b8R25

The logic in the front-page app for what is mentioned above is really just:  

```
if(flag) {// do something }
```

Unless you can see somewhere I could have some extra tests

ironsidevsquincy:
How is that flag being set?

ironsidevsquincy:
👍

GlynnPhillips:
@ironsidevsquincy I have moved all the `if` statements in one module that returns an object with everything needed. Hopefully this will tidy it up a lot as it got out of control when the scope started to creep and also make it easier to add to or remove in the future.

https://github.com/Financial-Times/next-front-page/pull/952/files#diff-6f2a864915fa2e3e43a9e29c852d0022R1

PR #953, opened 2016-07-22
remove z-index from horizontal rule above section columns

i-like-robots:
I'm not sure what the context of this line is so I'm not 100% sure if this change will mess other things up I'm unaware of. Currently it's the same colour as the background so it isn't visible by default (on the front page).

Anyway, the `z-index` is interfering with the mega menu and removing it appears to cause no harm. 

Or, should we remove it entirely?


ironsidevsquincy:
It's used to hide the first horizontal line in a section, so I'm guessing the z-index needs to be higher the the horizontal lines

i-like-robots:
Ah, tricky...

PR #962, opened 2016-08-01
Render paid posts in React via a client side proxy

adgad:
/cc @aintgoin2goa @andrewgeorgiou1981 

The current ad-served implementation of the paid post slot has many flaws. This is one attempt we're trying to come up with a better solution. The solution here makes a fetch request to an endpoint in the ads api which proxies the ad call and renders it as JSON. This JSON is then used to render a client-side react component in the paid post slot. All behind the `nativeAdsRendering=dfpProxy` flag.

(Note some of this, especially the template is currently still hacky and needs refactoring at some point).


andrewgeorgiou1981:
👍

PR #963, opened 2016-08-03
Server rendered native ads behind flag

adgad:
@aintgoin2goa @i-like-robots @andrewgeorgiou1981 
* Flag option to render the paid post on the server, polling the paid-post endpoint
* Remove paid-post template (well, leave it as a placeholder for now for client rendering option) and use n-section/card.


aintgoin2goa:
👍

config/pages.js
```diff
@@ -83,5 +84,18 @@ const getPages = (flags, region) => {
 export default (pageId, data, flags, region) => {
 	const sectionsData = sectionData(data, flags, region);
 	const pages = getPages(flags, region);
-	return pages[pageId].map(sectionId => getSection(sectionId, sectionsData[sectionId], flags, region));
+	return pages[pageId].map(sectionId => {
+		const sectionData = sectionsData[sectionId];
+		if(sectionData.content && Array.isArray(sectionData.content)) {
+			sectionData.content = sectionData.content.map(item => {
+				if(['Article', 'LiveBlog', 'FastFt'].indexOf(item.type) >= 0) {
+					return contentModel(item, { useCase: 'article-card' });
```
andygnewman: may also want to use the `excludeTaxonomies: true` use case - this was something that was quite specific to the front page and just filters out some tags (regions, people, organisations) from the primary tag computation as these are most likely to be repeated in the headline.

andygnewman:
Just the one comment about an n-content-model option that should be used for the front page, otherwise LGTM.

PR #965, opened 2016-08-05
Fix related stories where originating from primaries

andygnewman:
cc @adgad 

- get url property from graphql-api for primary theme and section
- bump n-content-model to get fix for passing items through in tag transform


adgad:
👍

PR #968, opened 2016-08-09
Add hack to allow headline testing temporarily

tom-parker:
cc @ironsidevsquincy @GlynnPhillips 

Headline data needs updated before merge


ironsidevsquincy:
👍

PR #971, opened 2016-08-10
WIP: Add targeting to client side paid post option

adgad:
Will finish this off on fri - needs a change to oAds to fire the oAds.initialised event, and probably could do with some validation of the data in the ad to ensure all is as we expect.

@VladDubrovskis @andrewgeorgiou1981


andrewgeorgiou1981:
👍

PR #975, opened 2016-08-15
Remove unsuccessful native ad code and use new flag

adgad:
Will make switching over easier. @VladDubrovskis @andrewgeorgiou1981 @aintgoin2goa


andrewgeorgiou1981:
👍

PR #976, opened 2016-08-15
Videos in top stories

leggsimon:
@i-like-robots @aintgoin2goa 

Related https://github.com/Financial-Times/n-card/pull/156 and https://github.com/Financial-Times/n-section/pull/85


config/layouts/headlines-video-story.js
```diff
@@ -0,0 +1,66 @@
...
+								components: [
+									{
+										type: 'Content',
+										size: 'large-no-image',
+										modifier: 'shorter',
+										standfirst: {
+											show: { default: true }
+										},
+										related: {
+											show: { default: true, S: false, M: false, L: true },
```
ironsidevsquincy: don't need the `M: false` (it stays at the last value set until changed)

config/layouts/headlines-video-story.js
```diff
@@ -0,0 +1,66 @@
...
+											colspan: { default: 12 }
```
ironsidevsquincy: is this needed? (think it's the default)

config/layouts/headlines-video-story.js
```diff
@@ -0,0 +1,66 @@
...
+							{
+								type: 'Column',
+								colspan: { default: 12, S: 6, L: 7 },
+								components: [
+									{
+										type: 'Content',
+										size: 'small',
+										modifier: 'video-big',
+										standfirst: {
+											show: { default: false, S: false, M: false, L: true, XL: false }
```
ironsidevsquincy: ditto, don't need `S: false, M: false`

config/layouts/headlines-video-story.js
leggsimon: Not sure to be honest, this template was copied from `headlines-picture-story.js` and tweaked. It has the same. Have removed it locally and it doesn't appear to have had any negative effect. Will remove.

adgad:
We need ads! Could you enable them in n-card?

config/layouts/headlines-video-story.js
```diff
@@ -0,0 +1,65 @@
...
+								colspan: { default: 12, S: 6, L: 7 },
+								components: [
+									{
+										type: 'Content',
+										size: 'small',
+										modifier: 'video-big',
+										standfirst: {
+											show: { default: false, L: true, XL: false }
+										},
+										image: createImageConfig({ default: 12, S: 6 }, { default: 'top' }, [322, 500])
```
ironsidevsquincy: Is the first param correct? At `L`, it's a 7 col in a 9 col, i.e. 7/12ths of 9/12ths

client/main.js
```diff
@@ -29,18 +29,7 @@ bootstrap(nUiConfig, ({ flags, mainCss }) => {
...
-					id: video.getAttribute('data-o-video-id'),
-					placeholder: true,
-					optimumwidth: video.getAttribute('data-o-video-optimum-width'),
-					classes: ['video'],
-					advertising: showVideoAdvertising,
-					source: 'brightcove'
-				});
-			})
-
+			oVideo.init();
```
aintgoin2goa: I'd like to know why this was done in the first place before we remove it?

config/layouts/headlines-video-story.js
```diff
@@ -0,0 +1,65 @@
...
+						type: 'Row',
+						components: [
+							{
+								type: 'Column',
+								colspan: { default: 12, S: 6, L: 5 },
+								components: [
+									{
+										type: 'Content',
+										size: 'large-no-image',
+										modifier: 'shorter',
```
aintgoin2goa: This is a really confusing parameter - what does it do?  It might need a better name, or at least a comment explaining what it does

config/layouts/headlines-video-story.js
leggsimon: I'm actually now thinking that that whole `image` line is necessary at all? It isn't used by `n-section` by the looks of it. It's only the `article` and `native-ad` models that use an image attribute

leggsimon: Unless it is used elsewhere?

ironsidevsquincy: So you don't show an image for this card?

leggsimon: Well we have the video's still but that is dealt with by o-video. We don't use the image component in n-card like an Article card does for example.

ironsidevsquincy:
👍

PR #977, opened 2016-08-15
Life & Arts component

andygout:
@ironsidevsquincy 

Have added a flag `frontPageLifeAndArts` with a couple of variants so designers can play around and work out layout/design tweaks.

Ultimately the variants are to be displayed at different times of the week, which is a separate piece of work.

screen shot 2016-08-15 at 15 50 01


config/pages.js
```diff
@@ -76,6 +76,9 @@ const getPages = (flags, region) => {
 		layout.splice(layout.indexOf(splicePosition), 0, 'brexit-secondary');
 	}
 
+	// Insert Life & Arts component directly prior to Video (if requisite flag is on)
+	if (flags.frontPageLifeAndArts) layout.splice(layout.indexOf('video'), 0, 'life-and-arts');
```
ironsidevsquincy: Do you also need to check `!== 'control'`?

config/layouts/life-and-arts-week.js
```diff
@@ -0,0 +1,50 @@
...
+		components: [
+			{
+				type: 'Column',
+				colspan: { default: 12, XL: 6 },
+				components: [
+					{
+						type: 'Content',
+						size: 'large-top-story',
+						image: createImageConfig(
+							{ default: 12 },
```
ironsidevsquincy: Don't think these are correct - you're saying the image is always 12 cols wide, but, for example, this component is 6 wide at XL

To see what images are being loaded, take the `max-width: 100%` style off `.n-image`

config/pages.js
andygout: When toggler is set to 'Control' the value of `flags.frontPageLifeAndArts` is `false` rather than `'control'`, so guessing the above should be sufficient?

config/layouts/life-and-arts-week.js
andygout: Slightly misunderstood this but get it now - will change.

config/pages.js
ironsidevsquincy: 👍

config/layouts/life-and-arts-weekend.js
```diff
@@ -52,7 +52,7 @@ export default flags => [
 					{
 						type: 'Content',
 						size: 'medium',
-						image: createImageConfig({ default: 12 }, { default: 'top' }, [12, 366, 449, 659]),
```
ironsidevsquincy: Just a general check that the widths are all correct (`659` seems very large, and `12`???)

ironsidevsquincy: (know it isn't this PR, but might be worth tidying up while here)

andygout: Whoops - 12 was a typo (312 intended).

The 659 images are falling at the extent of S layout (designs only displayed what was happening at onset of S, so don't think this had consideration) and have already agreed with designer that this needs revising as takes up an inordinate amount of the page, so those figures will be coming down.

Which bit needs tidying up?

ironsidevsquincy: If they're all correct, nothing!

ironsidevsquincy:
👍

PR #980, opened 2016-08-16
Dictate: Life & Arts layout + position by time of week

andygout:
cc @andygnewman @ironsidevsquincy 

I have positioned the weekend version of the component above Most Popular purely to indicate that it will need to be in a different position to the mid-week version and that this logic needs to be done at this stage; the actual positions are still TBC.

I seem to be duplicating files and wonder if better to: extract the code that decides if it is the weekend period and use in [`config/pages.js`](https://github.com/Financial-Times/next-front-page/blob/life-arts-component-time-dependent/config/pages.js#L92-L93) to decide the component's position and again in [`config/sections/index.js`](https://github.com/Financial-Times/next-front-page/blob/life-arts-component-time-dependent/config/sections/index.js#L41), giving that `isWeekendPeriod` value via an extra `opts` argument to dictate in a single [`life-and-arts` section file](https://github.com/Financial-Times/next-front-page/blob/life-arts-component-time-dependent/config/sections/life-and-arts-week.js#L5) which of the two `layoutId`s to apply (although it's possible that the two variants will eventually require differing section attributes and therefore separate files).

Once merged: 
- [ ] Change flag to on/off (currently has variants).


config/pages.js
```diff
@@ -76,8 +76,17 @@ const getPages = (flags, region) => {
 		layout.splice(layout.indexOf(splicePosition), 0, 'brexit-secondary');
 	}
 
-	// Insert Life & Arts component directly prior to Video (if requisite flag is on)
-	if (flags.frontPageLifeAndArts) layout.splice(layout.indexOf('video'), 0, 'life-and-arts');
+	// Dictate Life & Arts component layout and position according to time of week
+	if (flags.frontPageLifeAndArts) {
+		const date = new Date;
```
ironsidevsquincy: Interesting thread - http://stackoverflow.com/questions/3034941/new-myobject-vs-new-myobject
Guess it's probably best to always include parenthesis

config/sections/life-and-arts-weekend.js
```diff
@@ -0,0 +1,18 @@
+export default () => ({
+	id: 'life-and-arts',
```
ironsidevsquincy: Do they wan't to have different ids, so they can be tracked independently?

config/pages.js
andygout: Had forgotten to include and agree it is best having them present.

config/sections/life-and-arts-weekend.js
andygout: Good point - have amended in the refactor.

ironsidevsquincy:
👍

PR #992, opened 2016-08-22
Fix videos in top stories

leggsimon:


server/libs/section-data.js
```diff
@@ -20,14 +20,24 @@ const getLayoutHint = (data, flags) => {
...
+			layoutHint
 	}
 
 	return 'standard';
 };
 
+const secondItemIsAVideo = (data) => {
+	return data &&
+		data.items &&
+		data.items[1].isVideo &&
```
ironsidevsquincy: probably want to check there's at least two items too

server/libs/section-data.js
```diff
@@ -20,14 +20,24 @@ const getLayoutHint = (data, flags) => {
...
+		data.items[1].videoId
+}
+
 const getTopStoriesData = (data, flags = {}) => {
-	const layoutHint = getLayoutHint(data, flags);
+	let layoutHint = getLayoutHint(data, flags);
```
ironsidevsquincy: ?

aintgoin2goa:
LGTM (but I don't really understand what it is)

PR #997, opened 2016-08-23
stop using res.cache() method

wheresrhys:


wheresrhys:
@aintgoin2goa FYI, this PR is mainly about an API change, but also reduces stale-while-revalidate time, which might help with the stale content issues people have been seeing

aintgoin2goa:
oh, excellent!  👍

PR #1000, opened 2016-08-24
log missing images

bjfletcher:
sample log output:

> level=warn event=FRONT_PAGE_GRAPHQL_POLLER_MISSING_IMAGE queryName=frontPageUS queryVariables={'region':'US','ukSectionUuid':'2836ebbe-cd26-11de-a748-00144feabdc0','ukSectionIdV1':'Ng==-U2VjdGlvbnM=','usSectionUuid':'362127c8-fbbd-11df-b79a-00144feab49a','usSectionIdV1':'Mg==-U2VjdGlvbnM=','euSectionIdV1':'OQ==-U2VjdGlvbnM=','asiaSectionIdV1':'TnN0ZWluX0dMX1N0YW5kYWxvbmVfMDA5MzQx-R0w='} missingImageItems="[{'index':2,'type':'Article','id':'0747004c-69c8-11e6-ae5b-a7cc5dd5a28c','title':'Turkish forces launch offensive in northern Syria against Isis','published':'2016-08-24T07:32:24Z','lastPublished':'2016-08-24T11:26:30Z','isPrimaryImageDefined':true,'primaryImageRawSrc':null}]"
```

### CODE REVIEW

**Comment by aintgoin2goa** [comment]
You might want to make it `level=error` as I think those are the only ones which are sent to splunk

**Comment by bjfletcher** [comment]
@aintgoin2goa https://financialtimes.splunkcloud.com/en-US/app/search/search?q=search%20level%3Dwarn shows `warn` logs

But you've got me thinking that maybe this should be `error` instead of `warn`...

**Comment by aintgoin2goa** [comment]
It's controlled per-app by SPLUNK_LOG_LEVEL env var which defaults to `warn` and I don't see anything in config-vars which overrides this

File: server/libs/graphql-poller.js [hunk 1]
```diff
@@ -26,6 +26,28 @@ let queryResults = {
...
+			queryVariables: JSON.stringify(queryVariables),
+			missingImageItems: JSON.stringify(missingImageItems.map(i => ({
+				index: i.index,
+				type: i.type,
+				id: i.id,
+				title: i.title,
+				published: i.published,
+				lastPublished: i.lastPublished,
+				isPrimaryImageDefined: !!i.primaryImage,
+				primaryImageRawSrc: i.primaryImage ? i.primaryImage.rawSrc : null
```
ironsidevsquincy: this will always be null, wont it?

server/libs/graphql-poller.js
```diff
@@ -26,6 +26,28 @@ let queryResults = {
 	}
 };
 
+
+const logMissingImages = (items, queryName, queryVariables) => {
+	const missingImageItems = items.slice(0, 10) // interested only in the top 10 items
+		.map((item, index) => Object.assign({ index: index }, item)) // zip results with index for reference
+		.filter(item => !item.primaryImage || !item.primaryImage.rawSrc); // find items without image
+	if (missingImageItems.length) logger.warn({
```
ironsidevsquincy: `{ ... }`  around this might make it easier to read

ironsidevsquincy:
warns are appropriate I think, it'll log warn and higher

server/libs/graphql-poller.js
```diff
@@ -26,6 +26,28 @@ let queryResults = {
...
+			event: 'FRONT_PAGE_GRAPHQL_POLLER_MISSING_IMAGE',
+			queryName: queryName,
+			queryVariables: JSON.stringify(queryVariables),
+			missingImageItems: JSON.stringify(missingImageItems.map(i => ({
+				index: i.index,
+				type: i.type,
```
ironsidevsquincy: is this needed?

server/libs/graphql-poller.js
```diff
@@ -26,6 +26,28 @@ let queryResults = {
...
+				id: i.id,
+				title: i.title,
```
ironsidevsquincy: needed?

server/libs/graphql-poller.js
```diff
@@ -26,6 +26,28 @@ let queryResults = {
...
+				published: i.published,
```
ironsidevsquincy: needed?

ironsidevsquincy:
Could even move the logging to graphql (so there's less levels from it to the source), check the response that comes back from ES and log if there's no main image

server/libs/graphql-poller.js
bjfletcher: only if `i.primaryImage` isn't defined, right?

ironsidevsquincy: unless I'm reading it wrong, either `primaryImage` or `rawSrc` wont exist (due to your above filter, line 33), so either way this will always be `null`

server/libs/graphql-poller.js
bjfletcher: I think `published` date and `title` are useful information to quickly identify if someone's screenshot of missing image matches this by looking at the title, and I hope the `published` date can also help tell us what's going on. I'll remove `type`.

leggsimon:
@ironsidevsquincy @bjfletcher 
On moving it to GraphQL. If we check the response and it comes back with no main image, could we not cache it and keep attempting to get it from ES? (Or maybe re-ingest that article and try again?)

bjfletcher:
@leggsimon you read my mind, that was what I was thinking we could do... would like to understand what's going on first before we put in any solution/workaround

ironsidevsquincy:
Plus not all article do have images.

Flagging up articles without images would be a good start

bjfletcher:
@ironsidevsquincy I've limited the logging to just the first 10 top stories, I think at least the top 6 or 7 should have images, right?

ironsidevsquincy:
That does seem likely

leggsimon:
Sometimes we know whether a story _should_ have an image though. For example when the `layouthint` is `'standaloneimage` we know that the 2nd story should always have an image and generally it seems as though that's the one that gets flagged up most times, understandably.

ironsidevsquincy:
_should_, but that's no guarantee (have had articles in that slot without an image in the past)

PR #1001, opened 2016-08-24
remove unnecessary logging

bjfletcher:
as pointed out by @ironsidevsquincy


server/libs/graphql-poller.js
```diff
@@ -42,8 +42,7 @@ const logMissingImages = (items, queryName, queryVariables) => {
 				title: i.title,
 				published: i.published,
 				lastPublished: i.lastPublished,
-				isPrimaryImageDefined: !!i.primaryImage,
-				primaryImageRawSrc: i.primaryImage ? i.primaryImage.rawSrc : null
+				primaryImageRawSrc: i.primaryImage.rawSrc
```
ironsidevsquincy: but this will always be null???

ironsidevsquincy: in fact, this will potentially error (if `i.primarySrc` doesn't exist)

bjfletcher:
@ironsidevsquincy I really shouldn't be working today. :) Here's attempt #2...

server/libs/graphql-poller.js
```diff
@@ -42,7 +42,7 @@ const logMissingImages = (items, queryName, queryVariables) => {
 				title: i.title,
 				published: i.published,
 				lastPublished: i.lastPublished,
-				primaryImageRawSrc: i.primaryImage.rawSrc
+				primaryImage: JSON.stringify(i.primaryImage)
```
ironsidevsquincy: Right, but this will still _always be null_, unless I'm really missing something (which is totally possible)

You're filtering out items without a `primaryImage` or `primaryImage.rawSrc`, so what do you expect to output here?

bjfletcher: whether primaryImage is undefined, null, {}, { someRubbish: 1 } (without any rawSrc)... my idea was that JSON.stringify would tell us this

by the way, this logging has caught last night's missing image... result! :) now we know it's somewhere in between this and ES that is not working right

ironsidevsquincy: The query for `primaryImage` will only ever return the `rawSrcs`(https://github.com/Financial-Times/next-front-page/blob/master/config/queries.js#L35), i.e. it'll never have anything else in that object (you can also get `alt` and `src`, but we don't request them https://github.com/Financial-Times/next-graphql-api/blob/master/server/lib/types/content.js#L371-L397)

It'll always be null at this point, due to your filter

I think a good place to add logging would be https://github.com/Financial-Times/next-graphql-api/blob/master/server/lib/types/content.js#L175, then you can see exactly what the `mainImage` field is in the ES response we're getting

PR #1002, opened 2016-08-24
stop bypassing n-content-model

leggsimon:
Bypassing n-content-model was stopping the correct transforms happening to some image properties forcing some images not to display

@andygnewman @i-like-robots


matthew-andrews:
lol what +1 +1

PR #1003, opened 2016-08-25
Change: Single life & arts variant position by day and flag

andygnewman:
Objectives:
- single Life & Arts variant (5 story)
- position by default just above video at bottom of homepage
- reposition to above Opinion at weekend (based on day / hour) and with flag on (to be an AB test)

Default:
screen shot 2016-08-25 at 09 45 28

Weekend:
screen shot 2016-08-25 at 09 45 04


config/pages.js
```diff
@@ -86,10 +86,8 @@ const getPages = (flags, region, isWeekend) => {
 		layout.splice(layout.indexOf(splicePosition), 0, 'brexit-secondary');
 	}
 
-	if (flags.frontPageLifeAndArts) {
-		// Insert Life & Arts component into position dictated by time of week
-		layout.splice(layout.indexOf(isWeekend ? 'most-popular' : 'video'), 0, 'life-and-arts');
-	}
+	// Insert Life & Arts component into position dictated by time of week
+	layout.splice(layout.indexOf((flags.frontPageLifeAndArts && isWeekend) ? 'opinion' : 'video'), 0, 'life-and-arts');
```
ironsidevsquincy: Why the change? I would imagine the `frontPageLifeAndArts` flag dictates whether to show the section at all

config/sections/life-and-arts.js
```diff
@@ -1,8 +1,8 @@
-export default ({ isWeekend }) => ({
-	id: isWeekend ? 'life-and-arts-weekend' : 'life-and-arts-week',
```
ironsidevsquincy: do we not want different tracking depending on its position?

config/pages.js
andygnewman: Change of plan - default is for display at bottom - without flag, we're going to AB test the impact of putting it higher up at weekends.

However point taken that flag name is therefore incorrect, will create a new flag of `frontPageLifeAndArtsHigh` that will be the basis of the AB test.

config/sections/life-and-arts.js
andygnewman: Good point that on the scroll tracking we should probably have differentiation. Will pass through something for High / Low position.

ironsidevsquincy:
👍

PR #1005, opened 2016-08-25
hide markets data link on < L screens

leggsimon:
@i-like-robots 

hides markets data link on < L screens & refactors SCSS


i-like-robots:
LGTM

PR #1006, opened 2016-08-26
Re-introduce bypassing n-content-model

leggsimon:
So we can bypass n-content-model for content that are videos

@andygnewman


config/pages.js
```diff
@@ -102,7 +102,7 @@ export default (pageId, data, flags, region) => {
 		const sectionData = sectionsData[sectionId];
 		if(sectionData.content && Array.isArray(sectionData.content)) {
 			sectionData.content = sectionData.content.map(item => {
-				if(['Article', 'LiveBlog', 'FastFt'].indexOf(item.type) >= 0) {
+				if(['Article', 'LiveBlog', 'FastFt'].indexOf(item.type) >= 0 && !(item.isVideo && item.videoId)) {
```
andygnewman: I would only check item.isVideo.

if item.isVideo was true and item.videoId was false, it would still get passed to n-content-model, which it shouldn't because it's still a video.

andygnewman:
👍  now that isVideo is accurate.

Presume when Video becomes a type you'll just get rid of this `&& !isVideo`  because the first part based on types will take care of it.

leggsimon:
Yes exactly

PR #1011, opened 2016-08-30
Fail the GTG check

commuterjoy:
/cc @ironsidevsquincy secret


ironsidevsquincy:
👍  I wont tell anyone

server/init.js
```diff
@@ -51,8 +51,14 @@ startPolling();
 startBrexitLivePolling();
 startBrexitSecondaryLivePolling();
 
+// FIXME - temporary outage
+const hour = parseInt(process.env.simulated_outage) || 26;
```
leggsimon: Is this meant to be ` || 24` ?

commuterjoy: 26 just is an hour that doesn't exist, so it's switched off

PR #1012, opened 2016-08-31
Fix: app.js and init.js are now the right way round

keirog:
(Seeing the individual commits will hopefully give you a bit more confidence in thumbing)


ironsidevsquincy:
👍

PR #1013, opened 2016-09-01
Make: Landscape the default headlines layout

andygout:
@ironsidevsquincy @musthemag 

Following the negative reaction to the seldom-seen `Standard` layout last week, we are removing that option from the headlines layouts and making the default the landscape layout.

- [x] Approved by Editorial.
- [x] 'Standard' option removed from Methode CMS.
- [x] 'Standard' option removed from Next Flags API (and by association the Feature Toggle).


config/sections/headlines.js
```diff
@@ -9,9 +9,8 @@ const getLayout = (layoutHint) => {
 		case 'assassination':
 		case 'bigstory':
 			return 'headlines-big-story';
-		case 'standard':
 		default:
-			return 'headlines';
+			return 'headlines-landscape';
```
ironsidevsquincy: can remove the `landscape` case from this, right?

server/libs/section-data.js
```diff
@@ -10,7 +10,7 @@ const headlineStoryCount = layout => {
 		case 'bigstory': return 4;
 		case 'landscape': return 1;
 		case 'standaloneimage': return 2;
-		default: return 5;
+		default: return 1;
```
ironsidevsquincy: ditto

ironsidevsquincy:
Will need to update tests

config/sections/headlines.js
andygout: Reckon it's better to have `landscape` share with `default` so it's more explicit what is being returned when the layout hint is set to that option (it did this previously with `standard`).

ironsidevsquincy:
👍

PR #1019, opened 2016-09-06
Add US Election Midriff section

callumlocke:
This PR adds a section to the homepage for the US Election. It's more or less a clone of the Brexit one, which is coming off soon.

For now this new section only contains articles, but one of the five slots is to be replaced by a dynamic data graphic soon. The stories come from [this GraphQL source](https://github.com/Financial-Times/next-graphql-api/blob/899435d15057cbfe354dd49e487ffdc8adf526cb/server/config/sources.js#L50-L52).

image

Relevant flags:

- `frontPageUSElectionMidriffUK`
- `frontPageUSElectionMidriffInternational`

- [x] update [this line](https://github.com/Financial-Times/next-graphql-api/blob/b2c77cc755085a55b0261732beff3301e9656737/server/config/sources.js#L51) in the graphql API to use the Methode UUID of a list of US Election stories.
- [x] verify the section only appears with the right flag/edition combo, and is hidden otherwise
- [x] verify the 'follow' button is for the right stream

Also to do (but in another PR after this): add a dynamic data component in one of the slots, using similar technique to the 'brexit secondary interactive'


callumlocke:
@aintgoin2goa @keirog @i-like-robots could one of you check over this? I'd like to merge it in now if it's OK.

client/components/us-election-midriff/_us-election-midriff.scss
```diff
@@ -0,0 +1,64 @@
+@import 'o-grid/main';
```
ironsidevsquincy: is this needed?

ironsidevsquincy:
👍

config/sections/us-election-midriff.js
```diff
@@ -0,0 +1,29 @@
...
+	// isTab: true, // ?
+	layoutId: 'us-election-midriff',
+	trackable: 'us-election-midriff', // should be same as layoutId
+	trackScrollEvent: true,
+	// background: true, // ?
+	followButton: {
+		conceptId: 'N2UxNTM3MzItNWNlZC00MDc5LWI3ODUtYWNmZDA2YjE0MWE2-U2VjdGlvbnM=',
+		name: 'Brexit',
+		taxonomy: 'topics',
+		isVariant: data.flags ? data.flags.myFtAddCopyAbV1 : false
```
keirog: We‘re not using `isVariant` or the `myFtAddCopyAbV1` flag anymore, just so you know

client/components/us-election-midriff/_us-election-midriff.scss
callumlocke: no

config/sections/us-election-midriff.js
callumlocke: So I can just delete that line?

callumlocke:
Thanks

config/sections/us-election-midriff.js
keirog: Yep

PR #1021, opened 2016-09-06
switch silent mode off for o-video

i-like-robots:
https://github.com/Financial-Times/o-video/pull/34


i-like-robots:
CC @aintgoin2goa

PR #1023, opened 2016-09-07
Video next up

i-like-robots:
screen shot 2016-09-07 at 12 03 28

Displays the next up panel when the video ends if there is another video in the playlist. The panel will display for 5 seconds before being removed and triggering the next video or clicking it will skip the wait. On iOS (or other devices that trigger the `playsFullScreen` flag) nothing will happen.

This PR also simplifies some of the video play logic for a more conistent experience across devices and better delivery of ads.


client/components/video-scroller/player.js
```diff
@@ -77,6 +79,53 @@ function onVideoClick(e){
 	e.stopPropagation();
 }
 
+function onVideoEnd () {
+	if (state.playsFullScreen) {
+		return;
+	}
+
+	let index = state.playlist.currentIndex + 1;
```
ironsidevsquincy: `const`?

leggsimon:
Since there's no closing of full screen vids. I can close https://github.com/Financial-Times/next-front-page/pull/1025 ?

client/components/video-scroller/player.js
```diff
@@ -32,92 +34,101 @@ export function init(){
...
-	let placeholder = wrapper.querySelector('.o-video__placeholder');
-	video.style.width = placeholder.offsetWidth+'px';
-	video.style.height = placeholder.offsetHeight+'px';
-	wrapper.replaceChild(video, placeholder);
+function showUpNext (next) {
+	const container = state.elements.upNext = document.createElement('div');
+
+	container.className = 'video-scroller__next-up';
+
+	container.innerHTML = `
```
aintgoin2goa: This doesn't seem like a great idea - yeah, it's the easiest way to get it done but someone (possibly your future self) is gonna curse you for putting this here.

Could we at least move to the top of the file or something?

i-like-robots: Reckon it would be better to have this inactive in the DOM?

aintgoin2goa: Yes, that's much better I think

i-like-robots: Refactored this into the server-side templates

PR #1029, opened 2016-09-08
Add: Premium labels to myFT cards

andygout:
cc @andygnewman @bjfletcher 

Given the myFT components are handled client-side, it seems the quickest fix to apply the `premium` attribute natively here in `next-front-page` (given that is all we need) rather than the work involved in getting `n-content-model` to also work client-side, especially given long term plan is have the premium status supplied directly and not require a transform.

screen shot 2016-09-08 at 15 21 00


client/components/myft/myft.js
```diff
@@ -92,6 +92,10 @@ const handleResponse = (myFtContainerEl, flags, response) => {
 		.filter(filterEmptyConcepts)
 		.slice(0, 4);
 	concepts.reduce(filterDuplicateArticles, []);
+
+	// apply premium attribute to concept items
+	concepts.map(concept => concept.items.map(item => item.premium = /\/cms\/s\/3\//.test(item.webUrl)));
```
ironsidevsquincy: `map` here but nothing's being down with it's return value (same happening with the above `reduce`)

andygout: Are you saying we should be assigning to new variables rather than mutating in place?

ironsidevsquincy: Bigger picture, yes, defo (implicit mutating causes hard to track down bugs), but this in particular, you're using the map function which returns a 'mapped' array, but you're not doing anything it. So my question is why use map (as opposed to, say, `forEach`)

Regardless though, I'd change this to not do behind-the-scenes mutation

client/components/myft/myft.js
```diff
@@ -90,7 +90,12 @@ const handleResponse = (myFtContainerEl, flags, response) => {
 	const concepts = followed.concat(viewed, popularTopics)
 		.filter(filterDuplicateConcepts)
 		.filter(filterEmptyConcepts)
-		.slice(0, 4);
+		.slice(0, 4)
+		.map(concept => {
+			// add premium attribute to concept items
+			const items = concept.items.map(item => Object.assign({}, item, { premium: /\/cms\/s\/3\//.test(item.webUrl)}));
+			return Object.assign({}, concept, { items });
+		});
```
ironsidevsquincy: ideally this would come after the reduce, as there might be less items to iterate over, but the performance benefit vs the refactor isn't really worth blocking this

ironsidevsquincy:
👍

client/components/myft/myft.js
andygout: Will endeavour to refactor the `reduce` function so not mutating the concepts and at same time will put it ahead of this mapping.

PR #1030, opened 2016-09-08
Strengthen and make consistent data check for poller cache

andygnewman:
cc: @ironsidevsquincy 

A weaker poller data cache check vs content check on rendering is likely to have caused 500's.
This makes those checks consistent, so should always have valid data.
The existing graphQL health check should pick up when poller data is stale.

The improved logging for when the poller is unable to overwrite the cache (leading to stale content) should then aid further debugging, by indicating which data is missing.


server/libs/graphql-poller.js
```diff
@@ -67,12 +64,28 @@ const pollData = (query, variables, name, flags = {}) => {
...
+								if (results.hasOwnProperty(property)) {
+									return {
+										section: property,
+										items: property.items ? true : false,
+										itemsLength: property.items ? property.items.length : null
+									}
+								}
+							}
+						}
+					})
```
ironsidevsquincy: What are you trying to achieve here? stringify-ing a function seems a bit odd, plus you return after getting the first property. 

Maybe `map` over `Object.keys(results)`?

ironsidevsquincy:
👍

PR #1035, opened 2016-09-09
Add missing z-index to video scroller overlay container

i-like-robots:
Fixes an issue with winphone IE 11


shared/components/video-scroller/_video-scroller.scss
```diff
@@ -478,6 +478,7 @@
 	top: 0;
 	right: 0;
 	bottom: 0;
+	z-index: 10;
```
leggsimon: Did we not have the Origami `z-index` mixin that we were going to start using?

i-like-robots: I am a bad person

PR #1037, opened 2016-09-12
Put native ads behind a non-mvt flag ready to turn on

adgad:
Rename the flag(s) to a single `nativeAds` flag, and default to with images.

@VladDubrovskis @aintgoin2goa


VladDubrovskis:
👍

PR #1038, opened 2016-09-12
Carousel tweaks

i-like-robots:
screen shot 2016-09-12 at 12 42 01

- Updates colours
- Tweaks video card sizes to avoid pixel miscalculations
- Disables mouse events on 'inactive' cards


leggsimon:
LGTM

PR #1040, opened 2016-09-13
Amend: premium -> isPremium

andygout:
cc @bjfletcher @andygnewman


ironsidevsquincy:
Just thinking, could we not put this property in graphql?

andygout:
And do the transform from the `webUrl` in `next-graphql-api`?

ironsidevsquincy:
Just add a new property to the [Content](https://github.com/Financial-Times/next-graphql-api/blob/master/server/lib/types/content.js) model, `isPremium`, which does this check. Then anything using graphql will get it automatically

andygout:
Given our next step is to add the `isPremium` property directly to Elastic Search (allowing us to call that directly when acquiring data via GraphQL (rather than extracting it from the `webUrl` value)), it may well not be worth doing what would only be a temporary fix in the meantime.

leggsimon:
Streams don't use GraphQL though.

ironsidevsquincy:
Streams don't use GraphQL (though they should IMHO), but front page and myft do

Even when it's added to ES, the data will still come through GraphQL, it wont be calling it directly, so you need to put it in GraphQL anyway. If you do it now, when ES does supply it all you'll need to do is change GraphQL to get the value from the response (rather than inferring it from the webUrl), and there'll be no need to update clients (e.g. here). That's the beauty of GraphQL

andygout:
Am going to be adding `isPremium` to Elastic Search as my next job so seems to make most sense to do that and then make the amends to GraphQL (to get the value from the response directly). Otherwise am just adding an extra step of work in the meantime.

andygout:
^^^ As discussed, will make amends to GraphQL now given ES update could take considerable amount of time.

client/components/myft/myft.js
```diff
@@ -92,11 +92,9 @@ const handleResponse = (myFtContainerEl, flags, response) => {
 		.filter(filterDuplicateConcepts)
 		.filter(filterEmptyConcepts)
 		.slice(0, 4)
-		.map(concept => {
-			// add premium attribute to concept items
-			const items = concept.items.map(item => Object.assign({}, item, { isPremium: /\/cms\/s\/3\//.test(item.webUrl)}));
-			return Object.assign({}, concept, { items });
-		});
+
+	console.log('concepts ***: ', concepts);
```
leggsimon: Remove this

leggsimon:
LGTM 

You'll have to make sure you release the graphql change first (if you haven't already) and check it is available in the live playground just to make sure this doesn't all fall over

andygout:
Yup - merged yesterday: https://github.com/Financial-Times/next-graphql-api/pull/189

Just checked live playground now and all good.

PR #1042, opened 2016-09-13
Moves regional news from an ammit test to perm feature

GlynnPhillips:
Rather than have Ammit decide if a users is shown the regional news component we have moved this logic into the [CDN](https://github.com/Financial-Times/ft.com-cdn/blob/master/src/vcl/regional-news.vcl).

This set a new `FT-Regional-News` which is used within this change to display the correct component and also in the CDN to vary the cache.

New lists are being created by Mus so these sections will be editorially curated, which is why we have updated all the uuid's.


server/routes/front-page.js
```diff
@@ -37,7 +37,14 @@ export default region => (req, res) => {
 	components.PaidPost = PaidPost;
 	components.VideoScroller = VideoScroller;
 
-	let frontPageData = res.locals.flags.mockFrontPage ? 'mockFrontPage' : `frontPage${region}`;
+	const regionalNewsVariant = req.get('FT-Regional-News');
```
aintgoin2goa: I can't see a vary header being set anyway?  You need to set that to correctly vary the cache (unless someone has specifically told you not to for some reason?).

GlynnPhillips: It its handled by the CDN https://github.com/Financial-Times/ft.com-cdn/blob/master/src/vcl/regional-news.vcl#L22-L28

aintgoin2goa: Fair enough - didn't know that was the way we did things nowadays!

aintgoin2goa:
👍

PR #1044, opened 2016-09-14
Switching regional news to use the new list query

GlynnPhillips:


config/queries.js
```diff
@@ -154,7 +150,7 @@ const frontPage = `
 				... LiveBlogInfo
 			}
 		}
-		ukTopStories: page(uuid: $ukSectionUuid) {
+		ukTopStories: regionalNews(region: UK) {
```
ironsidevsquincy: How does the 'regional news' differ from the top stories list for a region? 
It's just we have a query for topStories which takes a region, and we have, say, ukTopStories, and I wonder if it's a bit confusing/subtle as to what the difference is

GlynnPhillips: Sorry I missed this comment, I have merged as its behind a flag and I want editorial to start looking at it.

I believe there are two problems here that will cause confusion and I want to tidy them up once we know everything is working correctly for editorial. 

1) My naming of `ukTopStories` as this should be more descriptive and inline with the rest of the naming around regionalNews. This is something I can tidy up quickly along with some removing of duplication that I know I can now remove since this is no longer a test.

2) The use of `Region` within the frontpage is very confusing as in most cases it should be`edition`. This is something that a little more intertwined across the front page and graphql so it will take me a little longer to review and update.

But to answer the question the difference between the two is that top stories list is passed a region that is really the edition and regional news is passed a region that is created based on a combination of your edition and location. This is why I created a separate query for them to keep that separation, I was cautious around using the region type as I thought that might cause confusion but when I did it I felt it was ok as long as the queries were different.

We could use the same query going forward and change the values of the Region type to be something like `UKEdition`, `IntEdition`, 'UKRegional'  but this is something I would like to have a chat about and do as a refinement

PR #1046, opened 2016-09-14
add system code to app

ifyio:
cc/ @commuterjoy @ironsidevsquincy

This is to ensure that the health check also spits out the system code


commuterjoy:
+1

PR #1049, opened 2016-09-15
Fix next up trying to find the incorrect o-video element

i-like-robots:
Oversight from the switch from to o-video v2.

We should wait for https://github.com/Financial-Times/o-video/pull/48 for the full effect


leggsimon:
👍

PR #1052, opened 2016-09-16
Adding layout builder

aintgoin2goa:
An attempt to refactor the way we build and alter the homepage layout, as so many people are dong different things with it at the same time.

@adgad 
@GlynnPhillips


config/layoutBuilder.js
```diff
@@ -0,0 +1,144 @@
...
+	withNextFtTour(flags){
+		if(flags.nextFtTour) {
+			this.insert('tour-tip').after('opinion');
+		}
+
+		return this;
+	}
+
+	withMarketsSection(flags){
+		console.log('withMarketsSection', flags.newMarketsNews)
```
VladDubrovskis: stray console.log 

Apart from this looks really good and is a lot more readable + makes more sense. Was looking in all of the code yesterday and was getting really confused/lost.

Apart from log LGTM

PR #1056, opened 2016-09-16
fix the position of ads behind a flag

VladDubrovskis:
/cc @adgad @aintgoin2goa


config/layoutBuilder.js
```diff
@@ -108,7 +110,7 @@ export default class LayoutBuilder {
 			(region.edition === 'US' && flags.frontPageUSElectionMidriffInternational)
 		) {
 
-			let insertPoint = this.hasSection('regional-news') ? 'regional-news' : 'mid-page-advert';
+			let insertPoint = this.hasSection('regional-top-stories') ? 'regional-top-stories' : 'mid-page-advert';
```
VladDubrovskis: @aintgoin2goa was testing my changes locally with "regional-news" and it was not behaving correctly, so updated my code and this line to use "regional-top-stories" which is a section name from pre-defined list. Just wanted to verify this is OK or if there is a case when there is 'regional-news'?

aintgoin2goa: Yeah sorry - my bad.  There's no such thing as regional-news

PR #1059, opened 2016-09-16
Upgrade frontpage app to Node 6

bjfletcher:


ironsidevsquincy:
Can we get rid of any of these? https://github.com/Financial-Times/next-front-page/blob/master/server/setup.js#L12-L17

i-like-robots:
did you manage to solve the webpack build issues you had earlier?

bjfletcher:
Darren, good idea. I'll do that in a separate PR.
Matt, yes, resolved.

PR #1063, opened 2016-09-19
Swap n-content-model and decorator

andygnewman:
@ironsidevsquincy This swaps out n-content-model (which replaced existing data) with n-content-decorator (which purely returns new / changed values).


config/pages.js
```diff
@@ -51,7 +51,7 @@ export default (pageId, data, flags, region) => {
 		if(sectionData.content && Array.isArray(sectionData.content)) {
 			sectionData.content = sectionData.content.map(item => {
 				if(['Article', 'LiveBlog', 'FastFt'].indexOf(item.type) >= 0 && !item.isVideo) {
-					return contentModel(item, { useCase: 'article-card', excludeTaxonomies: true});
+					return Object.assign(item, contentDecorator(item, { excludeTaxonomies: true }));
```
ironsidevsquincy: might be safer to create a new object, e.g.

```
return Object.assign({}, item, contentDecorator(item, { excludeTaxonomies: true }));
```

server/libs/section-data.js
```diff
@@ -46,7 +46,7 @@ const getTopStoriesData = (data, flags = {}) => {
 	let fastFt = data.frontPage.fastFT;
 
 	if(fastFt && Array.isArray(fastFt)) {
-		fastFt = fastFt.map(article => contentModel(Object.assign({}, article, { type: 'FastFt' }), { useCase: 'article-card'}));
+		fastFt = fastFt.map(article => Object.assign(article, contentDecorator(Object.assign({}, article, { type: 'FastFt' }))));
```
ironsidevsquincy: ditto

config/pages.js
andygnewman: Hmm, understand the reluctance to mutate the original object, but in this instance we're decorating the original object with the extra properties required for n-section / n-card. Which I would have thought would be best kept with the original object.

(Which is also how we've implemented it in stream and article). 

As background here's the original issue that led to changing from n-content-model to n-content-decorator https://github.com/Financial-Times/n-content-model/issues/23

ironsidevsquincy: But you're mapping over the content, so you can return a new object. Unless it's having some effect somewhere else in the code...? (which is the exact reason we should try and keep things immutable)

andygnewman: 👍

PR #1065, opened 2016-09-19
Rebuild/data model

aintgoin2goa:


leggsimon:
LGTM

getFrontPageData.js
```diff
@@ -0,0 +1,27 @@
...
+	body: JSON.stringify({
+		query,
+		variables
+	}),
+	headers: {
+		'Content-Type': 'application/json'
+	}
+})
+	.then(response => response.json())
+	.then(data => console.log(JSON.stringify(data, null, 2)));
```
leggsimon: Do we want this to `console.log`?

aintgoin2goa: Yeah - this is a script that gets the fixture used in the test - is run from `make fixturez

leggsimon: Ok coolz

PR #1066, opened 2016-09-20
a11y POC

lc512k:
~~Pending threshold PR in pa11y/ci https://github.com/pa11y/ci/pull/4~~

- Search for `Pa11y` in the CI logs under `provision` to see the output
- Checks a11y in CI without breaking the build, reports a11y failures
- Once all failures are fixed, the threshold option should be removed. This will cause any new a11y errors to break the build
- In the future we could look at moving a basic `.pa11yci.js` into `nht` that can be extended per project, along with a `nht a11y` recipe that would run pa11y without a threshold by default.

cc @rowanmanning @bjfletcher @adambraimbridge 

screen shot 2016-09-20 at 16 31 37


lc512k:
oops, sorry @rowanbeentje 😊

rowanmanning:
I've published version `0.3.0` of Pa11y CI, so you should be good to upgrade now 👍

.pa11yci.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
bjfletcher: How about renaming this file to `test/a11y.js`? I'm not 100% sure about making this a hidden file in root folder but if you like it like this, then I have no problem with that. :)

lc512k:
@bjfletcher its the default file name pa11y expexts

PR #1069, opened 2016-09-21
Clean up unnecessary babel transforms and update node-sass

i-like-robots:


i-like-robots:
CC @bjfletcher

bjfletcher:
Thanks, Matt!

Is `babel-plugin-add-module-exports` necessary having `babel-plugin-transform-es2015-modules-commonjs`?

bjfletcher:
BTW... `babel-plugin-transform-es2015-modules-commonjs` and `babel-preset-node6` are identical :) (The latter having just the first in its list of plugins.)

i-like-robots:
I don't think there's any harm going with a preset - I'm not sure of the reasons historically why we would choose _not_ to use them.

i-like-robots:
`babel-plugin-add-module-exports` is necessary (for now) because Babel did their own thing with ES6 modules. That means any module that tries to `require` an ES6 module has to add `.default`

PR #1070, opened 2016-09-21
don't ignore public directory

leggsimon:
According to @wheresrhys the whole `/public` directory shouldn't be ignored as n-webpack checks to see what files are in there using the `.gitignore` and by ignoring the whole `/public` directory it can't correctly check the existence of them.

cc/ @i-like-robots @aintgoin2goa


leggsimon:
Closing because @keirog beat me to it https://github.com/Financial-Times/next-front-page/pull/1073

PR #1071, opened 2016-09-21
Force FT-Poller to retry imeditaly if we recieve invalid data

aintgoin2goa:


leggsimon:
@ironsidevsquincy you know the other day we were talking about if an API (eg. next-video API) was flaky/receiving too much traffic and went down, would us constantly hitting it be a bad thing and potentially compound the issue. You weren't massively worried because the front page just polls every minute and so it wouldn't create tons more requests directly to the API. 

Would this negate that if it constantly retried every time it received bad data?

ironsidevsquincy:
Yeah, maybe there should be a counter in place, only try 3 times in a row or something, don't want to swamp the servers

aintgoin2goa:
Yeah, good point - will do that.  I will should probably also only do this on initial startup, after that we can just log an error and carry on.

aintgoin2goa:
@leggsimon @ironsidevsquincy It now retries a maximum of 3 times and only on startup.

Also made it a singleton so you can't accidentally spin up more pollers than you need

Ok?

PR #1072, opened 2016-09-21
Move the mobile ad after 6th story instead of 2nd

adgad:
@aintgoin2goa @VladDubrovskis @andrewgeorgiou1981 @gvonkoss 

Last one. Promise. Moves the 'threeAdProposition' mobile ad after the 6th article instead of after the 2nd.


andrewgeorgiou1981:
👍

PR #1077, opened 2016-09-22
Reduce the cache time

richard-still-ft:
This shouldn't really affect the cache hit rate because the volumes will be so high. Should be merged after we see the impact of the poller changes.

cc @matthew-andrews @wheresrhys @aintgoin2goa


matthew-andrews:
+1

PR #1079, opened 2016-09-22
Canary

aintgoin2goa:
Created a new "canary" version of the front-page, which pivots on an env var rather than a flag.

This new version is deployed to a separate heroku app.   Later on I will figure out how to proxy to it from the real front-page


server/init.js
```diff
@@ -1,2 +1,9 @@
 require('./setup');
-module.exports = require('./app');
+
+process.on('uncaughtException', err => {
```
aintgoin2goa: Appreciate this is a but I've some issues locally with uncaught exceptions and need to be sure it's ok on heroku

server/app-v2.js
```diff
@@ -0,0 +1,48 @@
...
+	withFlags: true,
+	withHandlebars: true,
+	withNavigation: true,
+	withAnonMiddleware: true,
+	withRequestTracing: true,
+	withBackendAuthentication: true,
+	hasHeadCss: true,
+	hasNUiBundle: true,
+	layoutsDir: path.join(process.cwd(), '/bower_components/n-ui/layout'),
+	// helpers: {}
```
leggsimon: remove this?

server/controllers/front-page.js
```diff
@@ -0,0 +1,64 @@
...
+	const renderParams = {
+		layout: 'wrapper',
+		region: region.edition
+	};
+	setupMarketsHeader(res.locals.flags, renderParams);
+	res.render('front-page-v2', renderParams);
+}
+
+function ukHomePage(req, res){
+	let edition = 'UK';
```
leggsimon: `const`

server/init.js
```diff
@@ -1,2 +1,9 @@
...
+	console.error(err.stack);
+});
+
+let app = process.env.CANARY ? require('./app-v2') : require('./app');
```
leggsimon: `const`

server/controllers/front-page.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
+let dataModel;
+
+const marketsDataTemplate =
+	`
+	<div class="markets-data-wrapper">
+		<div class="o-grid-container">
+			<div class="o-grid-row">
+				<div class="markets-data js-markets-data" data-o-grid-colspan="12" data-trackable="header | markets data">
```
i-like-robots: we can lose the `div.o-grid-row` element and `[data-o-grid-colspan="12"]` attribute

server/controllers/front-page.js
i-like-robots: const

server/controllers/front-page.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function ukHomePage(req, res){
+	let edition = 'UK';
+	const data = dataModel.ukHomePage;
+	return render(req, res, edition, data);
+}
+
+function internationalHomePage(req, res){
+	let edition = 'US';
```
i-like-robots: and here

server/controllers/front-page.js
```diff
@@ -0,0 +1,64 @@
...
+	const data = dataModel.internationalHomePage;
+	return render(req, res, edition, data);
+}
+
+module.exports = model => {
+	dataModel = model;
```
aintgoin2goa: It's used just above - but actually has been renamed to dataStore

server/app-v2.js
aintgoin2goa: That's to just to remind me to put them back later

PR #1080, opened 2016-09-22
Upsell premium to standard subscribers

bjfletcher:
This displays the (at present empty) premium promo component in place of two premium article cards, if the `frontPagePremiumUpsell` flag is turned on.

We'll do promo component's HTML and CSS in a separate PR. Just want to review the structure with @andygnewman before he goes on holiday.


server/routes/front-page.js
```diff
@@ -37,6 +38,7 @@ export default region => (req, res) => {
 	components.USElectionMidriffGraphic = USElectionMidriffGraphic;
 	components.PaidPost = PaidPost;
 	components.VideoScroller = VideoScroller;
+	components.PremiumPromo = PremiumPromo;
```
andygnewman: You could make this conditional upon the same flags - no point having the component if you're not rendering anything.

andygnewman:
Looking good.

PR #1082, opened 2016-09-24
Creating new Model , View Presenter structure

aintgoin2goa:
Also demonstrate it by building the headlines section.


ironsidevsquincy:
Have you got a general plan of how you see front-page/n-card/n-section changing, i.e. what the end goal of all these PRs will be?

aintgoin2goa:
@ironsidevsquincy 
The MVP in this case stands for Model View Presenter.  The goals are:

1. Make it easier for new devs and others in the FT to understand what's going on and how to make changes
2. Not have our layout engine make assumptions about the layout but to allow us to build anything design come up with
3. Make it easier to remove React at a later date, if we choose it
4. Make it easier to remove graphql as a later date, if we choose to
5. Work towards removing the need to babel in the app.

The plan is to remove n-section entirely, with the `Content` component currently moved into `n-card` (Ben hates this so this could change in the future).  N-card will remain unchanged.  We will use handlebars templates for layout with each section having a model, view and presenter - I think this is a common enough pattern to be easily understood by newcomers to the code.  React components will be wrapped in a helper which uses `React.createElement()` rather than JSX to enable the future removal of babel/register.  

I think/hope that returning to a more straightforward approach will make it much easier in the future.

ironsidevsquincy:
I think all the goals would be great to address.

Have you got any concrete examples of what things will look like, e.g. code

Also, assuming all this wont go live till after Tuesday

ironsidevsquincy:
Also, out of interest, what designs have been proposed that we are currently unable to achieve?

aintgoin2goa:
Err, yes - that is literally the point of this PR.  Admittedly there's a lot of code here, so check out any file called "presenter.js" "model.js" or anthing *.html to see the approach.  Also the "section" and "card" helpers

The whole thing is being built as a canary version which pivots on a env var so is viewable under it's own url and hopefully will figure out how to proxy to it.  I don't see it going live before Tuesday, though - will need some robust QAing first.

The new designs want to escape the idea of having the sections as horizontal stripes, which is pretty tricky using the current system.  Also the new video scroller has to be implemented as a custom react component as the idea of horizontal scrolling didn't really work either.  Brexit, US election and (I think) the new myFT tour are also custom components, so we've kind of reached the point when the layout system is being ignore anyway

ironsidevsquincy:
Yeah, but it's kinda tricky to understand what the outcome will be from a load of code. Even a simply diagram of how you see everything fitting together would be useful. Or maybe a session at the end of dev huddle

server/libs/card-config-presets.js
```diff
@@ -0,0 +1,53 @@
...
+configPresets.set('large-top-story', {
+	size: 'large-top-story',
+	standfirst: {
+		show: { default: true }
+	},
+	isMain: true,
+	image: createImageConfig({ default: 12, L: 9 }, { default: 'top', M: 'right' }, [300, 450, 517, 659])
+});
+
+configPresets.set('medium-image-top', {
```
leggsimon: maybe this should be `medium-top-image` to follow the `size-position-type` format that the others have

server/libs/create-image-sizes.js
```diff
@@ -0,0 +1,66 @@
...
+	}
+};
+
+export default (colspan, position) => {
+	let currentPagePadding;
+	let currentCardPadding;
+	let currentColspan;
+	let currentPosition;
+	let currentPageSize;
+	let currentSize;
```
leggsimon: Is it necessary to have all these `let`s? Can they not just be assigned below?

server/sections/headlines/model.js
```diff
@@ -0,0 +1,87 @@
...
+	return true;
+}
+
+function getItems(data, storyCount){
+	const items = data.topStories.items
+		.slice(0, storyCount)
+		.map(item => {
+			if(['Article', 'LiveBlog', 'FastFt'].includes(item.type) && !item.isVideo){
+				return Object.assign({}, item, decorate(item, { excludeTaxonomies: true }));
+			}else{
```
leggsimon: spaces

server/sections/headlines/model.js
```diff
@@ -0,0 +1,87 @@
...
+	constructor({data, flags}){
+		if(!checkData(data)){
+			return;
+		}
+
+		this.layoutHint = getLayoutHint(data, flags);
+		this.storyCount = headlineStoryCount(this.layoutHint);
+		this.items = getItems(data, this.storyCount);
+
+		//fixme - why does this get relatedContent and the code above doesn't??!!!
```
leggsimon: Does this need fixing before merge?

server/sections/headlines/presenter.js
```diff
@@ -0,0 +1,30 @@
+"use strict";
+const dataStore = require('../../data/data-store')();
+const ViewModel = require('./model');
+const Handlebars = require('handlebars');
+const fs = require('fs');
+const path = require('path');
+
+function loadView(name){
+	const templatePath = path.resolve(__dirname, `${name}.html`);
+	let template = fs.readFileSync(templatePath, {encoding:'utf8'});
```
leggsimon: const

server/sections/headlines/presenter.js
```diff
@@ -0,0 +1,30 @@
...
+	return Handlebars.compile(template);
+}
+
+const views = (function() {
```
leggsimon: Why is this a self-invoking function? Is there a better way? Only because this is very rarely used in our code

aintgoin2goa:
@ironsidevsquincy 
Does this help?
Basically, the data-store is responsible for  polling graphql, validating the response and holding it in memory.
The front-page controller renders out the main view file.
There is a section helper, which runs the required presenter, passing it everything in `res.locals`
The presenter grabs the data from the dataStore,  instantiates a viewModel, renders a handlbars template and returns the result.
The viewmodel takes care of massaging the data into the correct format for each section.
Finally, there is a `card` helper which renders out `n-card` with the correct props.
front page architecture - page 1

PR #1084, opened 2016-09-27
Implement: Premium promo upsell component style tweaks

andygout:
cc @bjfletcher 

Have reviewed with Oli. The one outstanding issue is that when the text wraps it goes underneath the green tick; desired effect is to align with first line of text, so will address that in a separate PR.


shared/components/premium-promo/_premium-promo.scss
```diff
@@ -1,29 +1,27 @@
...
-	color: oColorsGetPaletteColor('grey-tint5');
+	color: oColorsGetPaletteColor('claret');
 }
 
 .premium-promo__body {
-	@include oTypographySerifDisplay(s, $load-progressively: true);
+	@include oTypographyBody;
 	margin: 10px 0;
+	font-size: 16px;
+	line-height: 20.8px;
```
ironsidevsquincy: `.8`???

andygout: We were trying to create this, but maybe it is better to do it using a calculation as seems to have been done here?:

screen shot 2016-09-27 at 13 00 21

ironsidevsquincy: `oTypographyBody` mixin includes extra stuff that you're overriding immediately (https://github.com/Financial-Times/o-typography/blob/5dd0c90fbcd27cc51c4619c29924c590c5d79b57/scss/_body_general.scss#L18-L22), so why not just put in the `font` style directly

PR #1085, opened 2016-09-27
Optimise midriff graphic

callumlocke:
As per request from @wheresrhys (originally from Gadi), this PR is to separate the HTML and CSS for the US Election midriff graphic, so we can put the CSS into the head, to avoid a repaint.

Both CSS and HTML come from the same polled JSON object [here](https://ig.ft.com/us-elections/ec-forecast-component-2.json).

@wheresrhys can you help with this... I've put `{{{ usElectionMidriffGraphicStyles }}}` in the head (I think) but I'm not sure where these handlebars vars get set. So, somewhere, the `usElectionMidriffGraphicStyles` template var needs to be set to `<style>${pollerData.css}</style>`. For both international and UK pages. And only if the flag `frontPageUSElectionMidriffGraphic` is set.


server/libs/us-election-midriff-graphic-poller.js
```diff
@@ -1,7 +1,7 @@
 import logger from '@financial-times/n-logger';
 import Poller from 'ft-poller';
 
-const url = process.env.US_ELECTION_MIDRIFF_GRAPHIC_URL || 'https://ig.ft.com/us-elections/ec-forecast-component.json';
+const url = process.env.US_ELECTION_MIDRIFF_GRAPHIC_URL || 'https://ig.ft.com/us-elections/ec-forecast-component-2.json';
```
wheresrhys: how are you coordinating this code change with that env var in prod?

callumlocke: the env var is not set in prod. the only reason for the env var is to allow me to override it locally, so I can point it to another local server if necessary

callumlocke:
but did you see what I wrote in the description? it's not ready, I need to set a handlebars variable from somewhere

callumlocke:
OK as this is getting a bit hacky (taking the CSS and conditionally injecting it into the head depending on flags), I propose we close this without merging, and let @aintgoin2goa just port this section over to the new 'canary' architecture which is anticipated to go live next week.

wheresrhys:
I'd rather get it in before launch if possible. Will come and chat tomorrow. Even if the CSS is not behind a flag I think that's OK-ish because, if you've namespaced it properly (which it looks like you have) there'll be no damaging side-effects other than loading a little bit of extra redundant CSS when the flag is off. On balance I think I prefer that to a repaint.

PR #1087, opened 2016-09-27
Fix tooltip on US Election Midriff follow button

callumlocke:
This fixes the tooltip text shown when you hover over the follow button.


leggsimon:
Do you also need to update the `conceptId`?

callumlocke:
No the ID is correct.

PR #1089, opened 2016-09-28
only load myft graphql query on client side

wheresrhys:


config/queries-browser.js
```diff
@@ -0,0 +1,37 @@
...
+				... Basic
+			}
+			followed(limit: 4) {
+				... Basic
+			}
+		}
+	}
+`;
+
+export default { user };
```
ironsidevsquincy: no need for `default`

wheresrhys:
I might shelve this - has far less impact on file size than I anticipated, and there's a good chance we'll get rid of graphql soon anyway

PR #1090, opened 2016-09-28
Reduce cache time to 5s

richard-still-ft:
cc @commuterjoy @phamann I still think we have the volume of traffic for this TTL to not impact performance. I expect to see a fairly small hike in load to the front page app.

Thoughts?


server/routes/front-page.js
```diff
@@ -49,7 +49,7 @@ export default region => (req, res) => {
 		frontPage: getData(frontPageData)
 	};
 
-	res.set('Surrogate-Control', 'max-age=30, stale-while-revalidate=5, stale-if-error=86400');
+	res.set('Surrogate-Control', 'max-age=5, stale-while-revalidate=5, stale-if-error=86400');
```
phamann: Maybe up the `stale-while-revalidate` value?

PR #1097, opened 2016-09-29
Data store refactor

aintgoin2goa:


server/data/data-store.js
```diff
@@ -40,37 +43,31 @@ class FrontPageDataStore extends EventEmitter {
...
-			graphql.ukHomePage(this._proxy, 'ukHomePage', {query, flags}),
-			graphql.internationalHomePage(this._proxy, 'internationalHomePage', {query, flags})
-		]);
-	}
-
 	start(){
-		if(this._sources.has('graphql')){
-			this.startGraphQL();
+		const promises = [];
+		for(let [, source] of this._sources){
```
i-like-robots: 🤔 not sure what's happening here... is sources an array of arrays?

server/libs/presenter.js
```diff
@@ -1,4 +1,3 @@
-"use strict";
```
i-like-robots: Is this still required by node for proper `let` and `const` behaviour?

server/data/data-store.js
aintgoin2goa: It's a Map

PR #1098, opened 2016-09-29
Remove unused dependencies from package manifest

i-like-robots:
Should speed up builds by at least 3 or 4 seconds =]


i-like-robots:
Yeah - React for server side and Preact gets bundled on the client side so I understand it

PR #1101, opened 2016-09-29
Dismissable front-page tips

keirog:
- adds a cross to allow dismissal of the front-page tip, saved in localStorage
- only dismissable for enhanced users, as agreed
- adds a temp `nextFtTourTipDismiss` flag for testing
- not entirely re-usable for other apps, but shouldn’t be too much work to remedy that
- nothing whizzy

Little | Big
---|---
little | big

In action:
you-are-dismissed


client/components/tour-tip/index.js
```diff
@@ -0,0 +1,63 @@
+const superstore = require('superstore-sync');
+const HAS_DISMISSED_TIP = 'tour-tip-fp-dismissed';
+
+let dismissButton;
+let tipContainerEl;
+
+function hasLocalStorage () {
```
debugwand: we have this elsewhere, should at some point make it a common thing for testing. maybe add into superstore

client/components/tour-tip/index.js
```diff
@@ -0,0 +1,63 @@
...
+	superstore.local.unset(TEST_KEY);
+	return TEST_VAL === retrieved && superstore.isPersisting();
+}
+
+function hasDismissed () {
+	return Boolean(superstore.local.get(HAS_DISMISSED_TIP));
+}
+
+function removeTip () {
+	tipContainerEl.remove();
```
debugwand: any contained event listeners to tidy up?

client/components/tour-tip/index.js
```diff
@@ -0,0 +1,63 @@
...
+function dismissTip () {
+	superstore.local.set(HAS_DISMISSED_TIP, 1);
+	dismissButton.removeEventListener('click', dismissTip);
+	removeTip();
+}
+
+function setUpTipDismisser (appendDismisserTo) {
+	const appendDismisserToEl = tipContainerEl.querySelector(appendDismisserTo);
+
+	dismissButton = document.createElement('button');
```
debugwand: 👍

client/main.js
```diff
@@ -15,6 +16,11 @@ window.React = React;
 bootstrap(nUiConfig, ({ flags, mainCss }) => {
 	highlightDomPath();
 	tracking.scrollDepthComponents.init();
+	tourTip(flags, {
+		tipContainer: '#tour-tip',
```
debugwand: this isn't directly an issue with your code, but if we have >1 tips on a page, does that mean we have >1 identical IDs?

client/components/tour-tip/index.js
keirog: only the dismisser at the mo

client/main.js
keirog: The ID bit is just the n-section convention, where each section has its ID added to the HTML 🙊 . You‘re right that the dismisser code assumes only one tip per page, which I’m happy to fix. I was holding off refactoring until we needed it in other pages (where multiple tips are used). Not sure why I didn’t assume multiple tips from the start :embarassed:

PR #1103, opened 2016-09-29
Adding US election Midriff

aintgoin2goa:
@i-like-robots @leggsimon @callumlocke


i-like-robots:
Looks OK to me, @leggsimon ?

PR #1108, opened 2016-09-30
Quick fix for enhanced instant-alert buttons

keirog:
The front-page only loads myFT concept cards on the client side (unlike
other pages), so we need to initialise uiInstant after the cards have
been injected. Bit of a naff solution to be just running init again,
but it does the job.


keirog:
Yep, they both get those props on the first init in component-initializer. The non-instant version doesn’t have the same problem because it binds events without bothering to check for the existence of any forms.

PR #1110, opened 2016-09-30
US Election headpiece

callumlocke:
Adds a section to the top of the homepage for on-the-night coverage of the 2016 US elections.


client/components/us-election-headpiece/_mini-dashboard.scss
```diff
@@ -2,14 +2,49 @@
   box-sizing: border-box;
 
 	* {
-	  font: 16px MetricWeb, sans-serif;
+	  font: 16px/1 MetricWeb, sans-serif;
```
aintgoin2goa: No, the /1 refers to line-height

callumlocke:
@leggsimon @aintgoin2goa 

Is this OK to merge this in now? We want to QA it on canary.ft.com. It's all behind flags so won't be visible to users yet anyway.

server/helpers/section.js
```diff
@@ -11,6 +11,7 @@ function getPresenterPath (name){
 }
 
 fs.readdirSync(sectionsRootPath).forEach(name => {
+	if (name.charAt(0) === '.') return;
```
aintgoin2goa: What is this for?

server/sections/us-election-headpiece/view.html
```diff
@@ -0,0 +1,191 @@
...
+		</div>
+		<div id="us-election-headpiece-section-content" data-o-grid-colspan="12" class="section__column section__column--content">
+			<div class="section__column__inner">
+				<div class="o-grid-row">
+					<div class="column main-headline" data-o-grid-colspan="12">
+						<!-- TODO how to remove image from this card? -->
+						{{card items.[0] flags size="large-no-image"}}
+					</div>
+
+					<!-- MAIN STORY IMAGE - hack: repeat first card and hide text -->
```
aintgoin2goa: Is there anything we can do to help here?

callumlocke: The design includes an L-shaped 'card' that wraps around other elements, and it was difficult to make this work with just CSS floating etc. So our hack is to do two cards, and hide the text from the second one using CSS. This hack works fine but if you have a nicer solution let me know.

server/helpers/section.js
leggsimon: Maybe clearer to do it based on file extension?

server/sections/us-election-headpiece/view.html
leggsimon: Maybe this could be one of our new "teaser" types in the future @aintgoin2goa ? An L-shaped one.

client/components/us-election-headpiece/_us-election-headpiece.scss
```diff
@@ -0,0 +1,498 @@
...
+	// 	color: white;
+	// }
+	//
+	// .card__content--has-image {
+	// 	display: none;
+	// }
+	//
+	// .sub-column.left article.card:after {
+	// 	background: none;
+	// }
```
leggsimon: Can this all this be deleted?

client/components/us-election-headpiece/index.js
```diff
@@ -0,0 +1,24 @@
...
+			.then(resp => resp.json());
+	} else {
+		data = Promise.resolve({
+			AK: 'rep',
+			NY: 'dem',
+		});
+	}
+
+	Promise.all([svg, data]).then(values => {
+		const [map/*, mapData*/] = values;
```
leggsimon: `/*, mapData*/`??

server/sections/us-election-headpiece/view.html
callumlocke: The problem with an L-shaped card is that it wraps around other content. It either needs to be two different elements that can be placed independently in the page, or it would need to be some kind of special card that accepts arbitrary extra markup inside it (I guess that could be a Handlebars block helper).

client/components/us-election-headpiece/index.js
callumlocke: Later this is going to be updated to make use of that variable. I commented out to let eslint pass.

server/data/sources/us-election-mini-dashboard.datasource.js
```diff
@@ -0,0 +1,65 @@
...
+	// 			'rep_contested': 20
+	// 		},
+	// 		'house': {
+	// 			'dem': 261,
+	// 			'dem_pct': 30,
+	// 			'rep': 174,
+	// 			'rep_pct': 40
+	// 		}
+	// 	}
+	// }
```
callumlocke: it would be convenient for us to leave these commented lines here for now, so we can easily continue to test the display of many different variations of the data. between now and November 8 we are going to need to try lots of variations.

server/helpers/section.js
callumlocke: well there are other kinds of dotfile that can accumulate on some operating systems... I guess the most explicit way is to check if the filename is actually a directory. I'll do that

PR #1118, opened 2016-10-05
Reduce poll time, configure via env var and add timeout

adgad:
@richard-still-ft @commuterjoy @phamann @wheresrhys


richard-still-ft:
👍

PR #1125, opened 2016-10-07
Run the shared tour-tip dismisser

keirog:
[Related n-ui PR](https://github.com/Financial-Times/n-ui/pull/543)

Also updates the front-page-only dismisser, so that it doesn't interfere with the new, shared one


client/main.js
```diff
@@ -4,11 +4,12 @@ import OVideo from 'o-video';
 import highlightDomPath from './components/highlight-dom-path/highlight-dom-path';
 import marketsData from './components/markets-data/markets-data';
 import myft from './components/myft/myft';
-import tourTip from './components/tour-tip';
 import paidPost from './components/ads/paid-post';
 import nUiConfig from './n-ui-config';
 import * as videoScroller from './components/video-scroller';
+import tourTip from './components/tour-tip';
 const tracking = require('n-ui/tracking');
+const tipDismisser = require('n-ui/tour-tip/lib/dismiss.js');
```
andygout: `import` or `require` alone rather than mixture of two?

PR #1128, opened 2016-10-10
fast build cycle

bjfletcher:
Imagine a setup where you can change any code, and you see the change on the frontpage within one second. That's the dream, and it's not an impossible dream. :)

I have timeboxed this to just today. What I've done is:

- new `fast...` build & watch tasks
- build CSS using `node-sass` and a quick `split-css.js` script similar to Matt H's plugin for webpack
- watch & rebuild using `nodemon` for CSS and `webpack` for JS

it'd be nice to...

- use a watcher within the app, to reload CSS/JS files when changed, instead of restarting whole app
- add Live Reload capability


webpack.config.js
```diff
@@ -1,11 +1,16 @@
 const nWebpack = require('@financial-times/n-webpack');
 
+const entries = {
+		'./public/main.js': './client/main.js',
+};
+
+if (process.argv.indexOf('--no-css') === -1) {
```
leggsimon: Wouldn't this be true if `--no-css` wasn't there? Is that right?

ironsidevsquincy:
* does this produce sourcemaps?
 * when does the postcss step occur?

webpack.config.js
bjfletcher: yes - if `--no-css` isn't present then build CSS with webpack
if it is present, then don't build CSS with webpack (and instead use `node-sass` directly)

bjfletcher:
@ironsidevsquincy no sourcemaps & no postcss

bjfletcher:
@ironsidevsquincy sourcemaps added (though, I'm not sure how you use sourcemaps with inlined head CSS?) and postcss (autoprefixer) added. There's just IE8 left to do.

Interestingly, the resulting CSS files don't exactly match those built by Webpack. There's some slight differences in how the files are joined up (Webpack removes end of file newline, `node-sass` doesn't).  Some other differences I may have missed. This is because of `sass-loader` or `css-loader` doing some things that `node-sass` doesn't.

bjfletcher:
PR over in `n-makefile` https://github.com/financial-times/n-makefile

bjfletcher:
BTW, I need to work on some other things and so I'm hoping we can wrap this up soon.

i-like-robots:
Do we also need cssnano (as used by Webpack's CSS loader when the`minimize` option is set)?

bjfletcher:
@i-like-robots we could run that using `postcss` after `autoprefixer`, what do you think? Bearing in mind that I don't have much time to spend on all this. :) Be great if we can keep the jobs simple.

PR #1136, opened 2016-10-12
No need to babelify n-image and n-card

i-like-robots:
These components are now pre-transpiled when published on NPM


aintgoin2goa:
Please don't merge this yet!!  Will explain...

i-like-robots:
Is this 👌 now?

aintgoin2goa:
oh, yeah - should be fine now.  I can't actually remember what the problem was now...

PR #1140, opened 2016-10-13
delete lots of code

aintgoin2goa:


shared/components/video-scroller/_video-scroller.scss
```diff
@@ -1,253 +0,0 @@
-$video-scroller-spacing: 10px;
```
i-like-robots: where have these styles gone?

aintgoin2goa: Ok, put them back in the /client folder.  And found some more code to delete on the way.

PR #1144, opened 2016-10-13
US election headpiece improvements

callumlocke:
- simplifies markup and css in the headpiece section
- adds more clientside interactivity (live dashboard and live results map)
- renames data source to reflect that it now supports the live results map too


callumlocke:
@aintgoin2goa @leggsimon we'd like to merge this now, would you mind reviewing it?

You'll see there's still a few TODO comments, but I'd rather address these in separate, smaller PRs to avoid further big conflicts.

client/components/us-election-headpiece/_us-election-headpiece.scss
```diff
@@ -1,399 +1,363 @@
...
 		}
 
-		// target ALL 4 SMALL CARDS
-		&[data-size=small] {
-			padding-left: 13px;
-			padding-right: 4px;
+		&--2,
+		&--3,
+		&--4,
+		&--5 {
```
leggsimon: Maybe this would be nicer as `&:nth-child(n+2)`

client/components/us-election-headpiece/_us-election-headpiece.scss
```diff
@@ -1,399 +1,363 @@
...
 			}
-		}
 
-		.card__title,
-		.card__title-text {
-			color: $headpiece-text-color !important;
+			&--2,
+			&--3,
+			&--4,
+			&--5 {
```
leggsimon: `:nth-child(n+2)`

client/components/us-election-headpiece/_us-election-headpiece.scss
callumlocke: that wouldn't work as they're not all siblings. but agree this could be nicer, I'll add a new class to target these all smaller stories at once

client/components/us-election-headpiece/index.js
```diff
@@ -1,24 +1,310 @@
...
+	const panelElements = {};
+	for (const type of panelTypes.values()) {
+		panelElements[type] = headpieceElement.querySelector(`.mini-dashboard__panel--${type}`);
+	}
+
+	// set up the tabs
+	const tabElements = {};
+	let oTabs;
+	const ul = document.createElement('ul');
+	{
```
leggsimon: Any reason for the block statement?

client/components/us-election-headpiece/index.js
```diff
@@ -1,24 +1,310 @@
...
+		let i = 0;
+		for (const type of panelTypes.values()) {
+			const panel = panelElements[type];
+			if (!panel) continue;
+
+			const panelId = `mini-dashboard-panel-${type}`;
+			const heading = panel.querySelector('.mini-dashboard__panel-heading');
+
+			// add a tab for this panel
+			{
```
leggsimon: Any reason for the block statement?

client/components/us-election-headpiece/index.js
callumlocke: sorry, I merged without realising you hadn't finished reviewing.

Plain blocks used to be pointless, but now that `const` and `let` exist, they can be useful for locally scoping variables so they're only visible to code that actually needs them. Admittedly in this case these variables wouldn't cause any real harm if they leaked into the outer scope, but I'd rather not clutter the outer scope with irrelevant variables anyway. Makes it more resistant to possible future name clashes introduced by editing other parts of the file. (Personally I think there's also a secondary readability benefit to visually 'containing' a long and fiddly piece of code that does one job, and it even lets you collapse it in your editor like a function, but I guess that's subjective.)

PR #1147, opened 2016-10-14
Hide tour tips if user is anonymouse

keirog:


keirog:
Also adds flag checks that went missing in the front-page refactor

PR #1152, opened 2016-10-18
remove the backgorund code as we change it in oads now

VladDubrovskis:
/cc @adgad @andrewgeorgiou1981 @gvonkoss


andrewgeorgiou1981:
👍

PR #1158, opened 2016-10-19
Refactor data store

aintgoin2goa:
This is a complete re-write of the data layer.  Partly cos my initial implementation didn't make any sense.

Data is now separated out into buckets, eacb with it's own poller and validation etc.

This made it easy to split the graphql queries into chunks, which seems to make it faster and will make it easy to poll for topstories often but other stuff less often.

Should also make it easier in the future to try different architectures (ie doing the polling on workers instead of web nodes)


leggsimon:
Just my comments about splitting up the front page query into smaller components, also maybe it would be good to gitignore those really massive graphql responses?

server/data/sources/topstories.datasource.js
```diff
@@ -4,6 +4,6 @@ const GraphQLDataSource = require('./graphql.datasource');
 module.exports = class GraphQLTopStoriesDataSource extends GraphQLDataSource {
 
 	constructor ({region = ''}){
-		super({query, variables:{region:region}});
+		super({query, variables:{region:region}, interval:30000});
```
leggsimon: for consistency with `server/data/sources/otherhomepage.datasource.js`
```js
constructor ({region = ''}){
    const variables = {region};
    super({query, variables});
}
```

server/sections/video/presenter.js
```diff
@@ -5,8 +5,8 @@ const path = require('path');
 
 const view = presenter.loadView(path.resolve(__dirname, 'view.html'));
 
-module.exports = function videoSectionPresenter ({flags}){
-	const data = dataStore.retrieve('otherHomePage');
+module.exports = function videoSectionPresenter ({flags, edition}){
+	const data = dataStore.retrieve(`${edition}OtherHomePage`);
```
leggsimon: ditto

server/sections/technology/presenter.js
```diff
@@ -9,8 +9,8 @@ const views = {
 }
 
 
-module.exports = function technologyPresenter ({flags}){
-	const data = dataStore.retrieve('otherHomePage');
+module.exports = function technologyPresenter ({flags, edition}){
+	const data = dataStore.retrieve(`${edition}OtherHomePage`);
```
leggsimon: ditto

server/sections/premium/presenter.js
```diff
@@ -5,8 +5,8 @@ const path = require('path');
 
 const view = presenter.loadView(path.resolve(__dirname, 'view.html'));
 
-module.exports = function premiumPresenter ({flags}){
-	const data = dataStore.retrieve('otherHomePage');
+module.exports = function premiumPresenter ({flags, edition}){
+	const data = dataStore.retrieve(`${edition}OtherHomePage`);
```
leggsimon: `edition` should probably be `region`? (in fact in all the presenters)

server/sections/regional-news/presenter.js
```diff
@@ -24,7 +24,7 @@ module.exports = function regionalNewsPresenter ({flags, region, edition}){
 	const data = dataStore.retrieve('regionalNews');
 	const topStoriesData = dataStore.retrieve(`${edition}TopStories`);
```
leggsimon: ditto

server/sections/regional-news/presenter.js
```diff
@@ -24,7 +24,7 @@ module.exports = function regionalNewsPresenter ({flags, region, edition}){
...
 	const usElectionData = dataStore.retrieve('usElection');
-	const opinionData = dataStore.retrieve('otherHomePage').opinion;
+	const opinionData = dataStore.retrieve(`${edition}OtherHomePage`).opinion;
```
leggsimon: ditto

server/sections/premium/presenter.js
aintgoin2goa: Ideally graphql should be updated to use Edition not region - and to not use "US", which is misleading as it's not us at all.

ironsidevsquincy:
We'll sort out Edition/Region-gate in the new api (and I guess the old one too while we're maintaining both)

callumlocke:
this will probably conflict with my PR (#1144), but not too badly, so go ahead and merge this when you're happy with it and I'll rebase.

PR #1159, opened 2016-10-19
Move premium section to between ad and myFT

andygnewman:
@mmockett has confirmed location (between ad and myFT section)

This is behind a flag (frontPagePremiumSection) and will be subject to an A/B test.

screen shot 2016-10-19 at 18 23 28


leggsimon:
👍

PR #1163, opened 2016-10-20
Small layout fixes 

i-like-robots:
…to big story layout, us election midpiece and horizontal hairlines.

screen shot 2016-10-20 at 15 23 04

😎


leggsimon:
:+1:

PR #1194, opened 2016-10-28
Moves fastFT to the horizontal slot at the top

GlynnPhillips:
Initial work to move fastFT to the top slot with stories positioned side
by side.

In its current state, this is not final and has just been created as a
prototype for user testing.

This will sit behind the `horizontalFastFT` flag

screen shot 2016-10-28 at 14 05 10


server/sections/fastft/model.js
```diff
@@ -1,8 +1,21 @@
...
-module.exports = class FastFTViewModel extends ViewModel{
+function formatDate (data) {
+	let items = data.fastFT;
+
+	items.forEach((item) => {
+		item.dateText = ODate.format(item.lastPublished, 'datetime')
+	})
 
+	return data.fastFT.slice(0,4)
+}
```
bjfletcher: all looking great... this block of code can perhaps be turned into:

```
return data.fastFT
  .slice(0, 4)
  .map(item => Object.assign({}, item, { dateText: ODate.format(item.lastPublished, 'datetime') }));
```

PR #1195, opened 2016-10-28
Don't duplicate work done by n-logger

callumlocke:


server/data/bucket.js
```diff
@@ -42,8 +42,8 @@ class DataBucket extends EventEmitter {
 		this.retry();
 	}
 
-	onError (err){
-		log.error({event:'DATASOURCE_ERROR', error:err.message, stack:err.stack.replace(/\n/g, '; ')});
+	onError (error) {
+		log.error({ event:'DATASOURCE_ERROR', error });
```
leggsimon: Curiously, would this try and call the argument `error` on `log` as though it was a function? would `err` be better?

ironsidevsquincy: If 'understands' `error`s if they're a separate argument, so

```log.error({ event:'DATASOURCE_ERROR' }, error);```

PR #1211, opened 2016-11-02
Use the n-ui myft button instead

debugwand:


aendrew:
This breaks the CSS for the headpiece and midriff.

debugwand:
(css issue was old local n-ui)

PR #1218, opened 2016-11-03
US election headpiece tweaks, round six

aendrew:
Mostly IE fixes.


client/components/us-election-headpiece/_mini-dashboard.scss
```diff
@@ -370,6 +370,8 @@
 
 	&__progress-indicator {
 		width: 2em;
+		height: 2em;
+		z-index: 99;
```
callumlocke: any reason this needs to be 99? what is the value that it needs to beat?

client/components/us-election-headpiece/index.js
```diff
@@ -169,7 +169,7 @@ export default () => Promise.resolve().then(() => {
 			i++;
 		}
 
-		oTabs = Tabs.init(miniDashboard, {disableFocus: true})[0];
+		oTabs = Tabs.init(miniDashboard, {disablefocus: true})[0];
```
callumlocke: if this is due to a bug in o-tabs, could you file on issue report over there?

also to be safe, maybe set **both** property names here, in case it gets fixed in o-tabs, i.e. `{disablefocus: true, disableFocus: true}`

aendrew: @callumlocke It's not a bug, it's like that in the documentation. Why they've decided to not use camelcase in this instance is beyond me.

callumlocke: it's ok, I opened an issue https://github.com/Financial-Times/o-tabs/issues/40

client/components/us-election-headpiece/_mini-dashboard.scss
aendrew: I realised that was pretty high and might interfere with modals and whatnot so reduced it to 5.

PR #1224, opened 2016-11-04
Election headpiece battleproofing

aendrew:
This does a few things:

a. Makes the tabs not appear broken if fonts don't load (Particularly problematic on Windows Phone)
b. No longer prevents the app from coming up in the event that the Heroku service supplying the election data goes down
c. Allows the toggling of the link to results based on data from the JSON file

I may add a slight restyling of one of the charts to this; please hold off on merging for now.


kavanagh:
LGTM 👍

PR #1226, opened 2016-11-07
🆙 Replacing US election headpiece endpoint URL with LIVE url

aendrew:
This sets the endpoint URL to the live election results app.


kavanagh:
LGTM 👍

PR #1228, opened 2016-11-07
🆙 Election headpiece: go-live tweaks

aendrew:
Myriad election headpiece tweaks:

- Increase image sizes for L/XL breakpoints
- Remove extra margin above 2x2 grid on L+ breakpoints when no interactive element


kavanagh:
LGTM 👍

PR #1229, opened 2016-11-07
Fix for undefined section appearing

adgad:
@leggsimon @aintgoin2goa @TheoLeanse 
screen shot 2016-11-07 at 12 30 26


GlynnPhillips:
👍

PR #1233, opened 2016-11-07
⚡️ Election headpiece: adds polltracker widget

aendrew:
At the request of an editor, I've added the graphic from the midriff to the headpiece.


callumlocke:
thanks for putting this together so quickly @aendrew

PR #1234, opened 2016-11-07
⚡️ Election headpiece: redirect markets tab to markets.ft.com

aendrew:
Currently the markets tab redirects to the results; it should redirect to markets.ft.com.


kavanagh:
👍

PR #1238, opened 2016-11-08
🐛 Redid election IE <=9 condit tags to use tables

aendrew:
The last attempt to fix the election headpiece in IE <=9 had issues, so this is another attempt, this time using tables. Works surprisingly well...


kavanagh:
I suppose we can live with this as it's only for a couple of days 👍

PR #1241, opened 2016-11-08
🐛 Election headpiece: Fixes reflow after tabs load

aendrew:
This adds a default height to the tab dashboard container so that the stories beneath don't jump around once a tab loads.


kavanagh:
:shipit:

PR #1244, opened 2016-11-09
⚡️ Election: Sets more aggressive polling intervals

aendrew:
Previously "timeout" wasn't set in the correct config property and refreshInterval was the default (60000). This sets timeout to 8s and refreshInterval to 20s.


kavanagh:
👍 We can back off a bit in a couple of hours when the result is done.

PR #1246, opened 2016-11-09
⚡️ Election: Reverts timings and sets results to off by default

aendrew:
Less aggressive polling and results set to off by default.


aendrew:
Reviewed by @kavanagh.

PR #1247, opened 2016-11-09
⚡️ Election: Reverting mini polltracker code

aendrew:
This removes the mini polltracker code added from the midriff on Monday.


kavanagh:
👍 Grid rid!

PR #1257, opened 2016-11-09
Make election data static

callumlocke:


aendrew:
@callumlocke I think the initial emit needs to be done in the `start()` method, or else it's too early and nothing comes up. See added commit.

aintgoin2goa:
@callumlocke @aendrew Do you still need this?

aendrew:
@aintgoin2goa Nope, I think we can close this.

PR #1259, opened 2016-11-10
➕ Adds: Trump midriff

aendrew:
This is a WIP PR intended to:

- A query to the Trump ID (`TnN0ZWluX1BOXzIwMDkwNjIzXzE2MTY=-UE4=`)
- Render a midriff with Trumpian stories

Currently it bails before the coming up because the GraphQL query seems to be failing. Any input into why would be most appreciated.


aendrew:
This is meant to still feed from a Methode list; see #1260.

PR #1260, opened 2016-11-10
🔸 Replaces us-election-midriff with donald-trump-midriff

aendrew:
Basically just renames a bunch of stuff. Also removes a lot of the election midriff graphics stuff.

screen shot 2016-11-10 at 15 59 28


server/sections/regional-news/presenter.js
```diff
@@ -32,7 +32,7 @@ module.exports = function regionalNewsPresenter ({flags, region, edition}){
 	const otherItems = concat([
 		topStoriesData.topStories,
 		usElectionData.usElectionHeadpiece,
-		usElectionData.usElectionMidriff,
+		// usElectionData.trumpMidriff,
```
kavanagh: Why has this been removed?

aendrew: @kavanagh I figured I'd just replace any instance of usElectionMidriff, that was one of the ones I wasn't sure about. I'm not sure `usElectionData.usElectionMidriff` is there anymore given that I've replaced it elsewhere, though I can revert that pretty easily.

kavanagh: I don't think I should review this. I'm not sure what that does. 

How does this pertain to the Regional News section? Perhaps this is something to do with deduping.

kavanagh:
Not sure we needed to rename everything. But I suppose it's done now.

server/sections/regional-news/presenter.js
aendrew: @kavanagh That'd be my guess. Fair point though, will get somebody on the next-front-page team to take a look.

server/data/graphql-queries/parts/trump-midriff.js
```diff
@@ -0,0 +1,15 @@
+// This is still called "usElectionMidriff" because an update is
+// needed at:
+// https://github.com/Financial-Times/next-api/blob/master/server/graphql/apis/us-election-midriff.js
+
+
```
andygout: Change has been made in `next-api` so can remove these comments + blank lines.

andygout:
Cannot be merged until `next-api` has correct list value got [`trumpMidriff`](https://github.com/Financial-Times/next-api/blob/master/server/config/ids.js#L38) so making as 'changes requested' until then.

server/sections/regional-news/presenter.js
GlynnPhillips: CORRECT 🎉 This seems right to me

PR #1264, opened 2016-11-10
⚡️ Tweak: Replaces US election feed in headpiece with Trump stream

aendrew:
Also does some tweaks to the progress indicator if there is only one tab.


kavanagh:
LGMT 👍

PR #1265, opened 2016-11-10
updated stacking widths

gvonkoss:
@adgad @keirog


views/partials/teaser-collections/opinion.html
```diff
@@ -6,13 +6,13 @@
 	</div>
 
 	<div data-o-grid-colspan="12 M6 L4">
-		{{#slice items offset=1 limit=4}}
+		{{#slice items offset=1 limit=3}}
 			{{>n-teaser/templates/light mods=(array 'small')}}
```
adgad: I think this needs to be standard too?

PR #1280, opened 2016-11-16
Direct: All myFT traffic to next-api

andygout:
cc @ironsidevsquincy


ironsidevsquincy:
Can tidy this too - https://github.com/Financial-Times/next-front-page/blob/29a660669661f8eb4a2a069706417ce1956f06a6/client/components/myft/myft.js#L84-L85

PR #1282, opened 2016-11-16
pa11y

lc512k:
Turns on pa11y for the front-page

More work remains to run the tests on mock data, rather than on whatever's on the front page that day.

@andygout @keirog @GlynnPhillips


keirog:
👍

PR #1294, opened 2016-11-21
Hide: Components w/null next-api results (rather than error)

andygout:
@GlynnPhillips @aintgoin2goa @ironsidevsquincy 

If a query to `next-api` returns `null` for a given section, it will not render, rather than throwing a page error.

- [x] US Election headpiece
- [x] FastFT
- [x] Donald Trump Midriff
- [x] Regional news (requires both Regional and Search (for recent stories) query to return null)
- [x] Opinion
- [x] Video
- [x] Highlights
- [x] Most Read
- [x] Technology
- [x] Market News
- [x] Life & Arts

### Work on separate branch (https://github.com/Financial-Times/next-front-page/pull/1298):
- Handling if non-Top Stories `next-api` queries error, again by not displaying component(s) that rely on given query.


server/sections/donald-trump/presenter.js
```diff
@@ -20,5 +20,5 @@ module.exports = function usElectionMidriffPresenter ({edition, flags}){
 	const data = {homepage: dataStore.retrieve('usElection')};
 	const model = new ViewModel({data, flags});
 	const view = flags.nTeaserFrontPage ? views.teasers : views.cards;
-	return view(model);
+	return model.items.length ? view(model) : null;
```
andygout: Yup - have tested on all components listed above.

PR #1296, opened 2016-11-21
Additional: GraphQL health checks

andygout:
cc @GlynnPhillips @ironsidevsquincy @kavanagh @aendrew 

Re. US Election Headpiece - even though its flag is now turned off and given Editorial want to keep it for time being, does not seem like a bad idea to give it a health check until the code is removed.


aendrew:
Cheers, @andygout — good idea. :+1: 

Thanks for setting that up!

PR #1297, opened 2016-11-21
Teaser grid tweaks

adgad:
@keirog


client/main.scss
```diff
@@ -17,6 +17,12 @@ $output-critical: true !default;
 
 @include nUiCriticalStart();
 
+[data-o-grid-colspan~=XLoffset1] {
+	@include oGridRespondTo(XL) {
+		@include oGridOffset(1);
+	}
+}
+
```
keirog: Could use [`nUiGridOffset`](https://github.com/Financial-Times/n-ui/blob/master/grid/README.md#offset-columns) instead

PR #1298, opened 2016-11-21
Store: Empty object in data bucket on error

andygout:
cc @ironsidevsquincy @aintgoin2goa 

Dependent on this PR: https://github.com/Financial-Times/next-front-page/pull/1294

If one of the below queries fails, the components that rely on the query will not display.

- Fast FT
- US Election
- Regional News
- Other homepage (Opinion, Video, Highlights, Most Read, Technology, Markets, Life & Arts)

Will prevent scenarios like [this](https://github.com/Financial-Times/next-bugs/blob/master/incidents/2016-11-12%20The%20one%20were%20the%20front-page%20went%20pop.md) from happening.

### TO DO
- [x] Test query: Fast FT
- [x] Test query: US Election
- [x] Test query: Regional News
- [x] Test query: Other homepage
- [x] Ensure that front page does still fail if no Top Stories present.
- [x] Health check on `next-api` for 4xx errors to alert if failing queries are being made: [next-api/server/config/health-checks/app.js](https://github.com/Financial-Times/next-api/blob/master/server/config/health-checks/app.js#L6-L14) (although may need to amend threshold).


server/data/bucket.js
```diff
@@ -43,7 +43,8 @@ class DataBucket extends EventEmitter {
 	}
 
 	onError (err){
-		log.error({event:'DATASOURCE_ERROR', error:err.message, stack:err.stack.replace(/\n/g, '; ')});
+		log.error({ event: 'DATASOURCE_ERROR', name: this.name, error: err });
```
ironsidevsquincy: just pass the err through as a separate argument, logger takes care of pulling out the useful stuff

```log.error({ event: 'DATASOURCE_ERROR', name: this.name }, err);```

PR #1313, opened 2016-11-24
fix compact view

aintgoin2goa:
Hey, @aendrew you missed a bit...

PR #1314, opened 2016-11-24
Remove: Remaining US Election graphics code

andygout:
@aendrew Few more remaining bits relating to IG.


views/front-page-n-teaser.html
```diff
@@ -4,9 +4,6 @@
 	<link rel="alternate" href="https://www.ft.com" hreflang="x-default">
 	<meta name="dfp_site" content="home">
 	<meta name="dfp_zone" content="{{region.edition}}">
-	{{#if @root.flags.frontPageUSElectionMidriffGraphic}}
-		<style>{{{usElectionGraphicCss}}}</style>
-	{{/if}}
```
aendrew: I think this is removed in #1313.

andygout: That was from `views/front-page-compact.html`.

aendrew: Ah, right you are!

PR #1328, opened 2016-11-25
Hide fast ft at weekends

andygnewman:
cc: @ironsidevsquincy 

Proposition to not show FastFT component at weekends (ie. when none are being published).

This looks at the time stamp of the most recent article and decides whether or not to display determined by time difference vs current time (set to 4 hours).


server/sections/fastft/presenter.js
```diff
@@ -17,6 +19,14 @@ const views = {
 
 module.exports = function fastFTPresenter ({flags}){
 	const data = dataStore.retrieve('fastFt');
+	// if inactive do not show fastft
+	const mostRecentArticle = data && data.fastFt && data.fastFt[0];
```
ironsidevsquincy: _Could_ (if `data` can be null)

```const { fastFt: [mostRecentArticle] = [] } = (data || {});```

Though that might be harder to read

andygnewman: Like it.

PR #1342, opened 2016-12-01
add view toggle url

aintgoin2goa:


package.json
```diff
@@ -31,7 +31,8 @@
     "preact-compat": "3.4.2",
     "proptypes": "^0.14.3",
     "react": "^15.3.2",
-    "react-dom": "^15.3.2"
+    "react-dom": "^15.3.2",
+    "supertest": "^2.0.1"
```
leggsimon: Should be a `devDependency`?

aintgoin2goa: yes

PR #1362, opened 2016-12-06
Remove: Regional News health checks

andygout:
cc @GlynnPhillips 

As discussed, given the Regional News component requires de-duping of articles from two different sources, conducting health checks for the amount returned by each of those is not particularly helpful in comparison to knowing the number of articles post de-duping.

On a separate branch we will add logging for the final number of articles for the Regional News component and implement a health check to monitor this data instead.


GlynnPhillips:
💯

PR #1375, opened 2016-12-08
Add: Health check for Regional News model having no articles

andygout:
cc @aintgoin2goa @ironsidevsquincy @GlynnPhillips

As mentioned in this PR (https://github.com/Financial-Times/next-front-page/pull/1362), this is to implement a health check to check that Regional news has at least one article after combining sources and de-duping (given its nature of using multiple data sources, using a GraphQL query health check won't give an accurate picture of whether the component has insufficient data).


server/sections/regional-news/model.js
```diff
@@ -31,6 +32,10 @@ function getItems (data, storyCount, region, otherItems){
 		}
 	}
 
+	if (!items.length) {
+		metrics.count('sections.models.regional-news.no-articles.count', 1);
```
ironsidevsquincy: would it be better to histogram the actual number of items, then set up an alert for if that number is less than 5 (or whatever it is)?

andygout:
@aintgoin2goa has said that maybe this isn't the best use of a health check (more intended to show upstream services are responding as expected) and maybe we can just monitor and trigger alerts directly from Graphite. Thoughts?

ironsidevsquincy:
Healthchecks aren't just for checking upstream dependencies, they're to flag up if the app is in an unhealthy/bad state, which can obviously due to a number of different reasons. 
The question is, if there isn't enough stories, do we consider the app unhealthy? Should someone be notified? Seems yes to me, but don't know enough about what the repercussions are to know for sure

ironsidevsquincy:
If it's more an fyi, then yeah, set up a grafana alert and post to an editorial slack channel or something

andygout:
Ultimately the purpose of the (GraphQL) health checks is to alert if there is insufficient content to display for respective components, it's just the way Regional News is set up requires a different approach, but it should be treated with the same importance as the other components and likewise use a health check.

andygout:
Also think maybe it's worthwhile deploying the metrics in `server/sections/regional-news/model.js` before adding the health check in `server/config/health-checks/models.js` so that some metrics have time to get processed into Graphite (else alerts may get fired given we're listening for figures below a threshold).

server/config/health-checks/models.js
```diff
@@ -0,0 +1,20 @@
...
+	name: 'Models',
+	description : 'Presence of content for models that manipulate data post-GraphQL query',
+	checks : [
+		{
+			name: 'regional-news content',
+			severity: 2,
+			businessImpact: 'No Regional News section displayed on the front page.',
+			technicalSummary: 'Checks that there is at least one Regional News articles post combining sources and de-duping.',
+			panicGuide: 'Check the Next API service health checks (https://next-api.ft.com/__health).',
+			type: 'availability',
```
ironsidevsquincy: what is this type?

andygout: Oop, yeah, should be `graphiteThreshold`.

PR #1379, opened 2016-12-08
Remove parent trackable ID from top stories

keirog:
We have more granular events from individual top story slices, so we can remove the parent one at Moshe's request.


keirog:
@GlynnPhillips

PR #1380, opened 2016-12-08
Minor teaser tidy-up

keirog:
Renames all the `[name]--teaser.html` templates to be `[name].html`, now that teasers are the default.

Best viewed as individual commits.

@andygnewman


andygnewman:
cc: @keirog does this fix all the issues with turning the flag off?

PR #1386, opened 2016-12-09
Fix: GraphQL health checks

andygout:
cc @ironsidevsquincy @aintgoin2goa 

It seems the term `verify` is used (rather than `verifyKeys`) ([here](https://github.com/Financial-Times/next-front-page/blob/master/server/libs/health-checks/graphql.js#L17)).

Checking locally on https://local.ft.com:5050/front-page/__health, if the value of Technology's verify value is increased from 4 to 40 then the health check fails as expected (whereas it didn't prior to changes). 

Also sets the Premium verify value to correct format.


ironsidevsquincy:
lol. wonder if we're going to get a rash of alarms now

PR #1407, opened 2016-12-13
remove ads from the tab order

lc512k:
Remove ads from tab order and hide from screen readers.

Addresses issue `DAC-USER-KO-T2-01` from the accessibility audit report.

@GlynnPhillips @jkerr321


GlynnPhillips:
Have you looked at also adding `role="presentation"` there are a few test results I have read that suggest the combination of all of these give the best results but I read all of that a long time ago so can't remember all the details of it.

Here is an example http://john.foliot.ca/html5/aria-hidden/

lc512k:
I've added `role="presentation"` to all of them, @GlynnPhillips. Thanks for that link!

lc512k:
@GlynnPhillips wel need to get those attributes deeper into the hierarchy, it seems. We can still tab our way to the google iframe inside o-ads. It's even available to a screen reader :( 

Will think of a workaround tomorrow

screen shot 2016-12-13 at 20 43 28

lc512k:
We'd need to set tabindex="-1" directly on the iframe. Thought tabindex would be inherited by all the children :/

https://jsbin.com/hakuqamure/1/edit?html,output

lc512k:
this might do? https://github.com/Financial-Times/o-ads/pull/98

PR #1408, opened 2016-12-13
Update: GraphQL health checks

andygout:
cc @aintgoin2goa 

Values updated to represent number of items required for respective components.

N.B. Top Stories requires 15 items, and `topStories`' first item is the same as the single `topStory` item, and so this needs to be accounted for.


ironsidevsquincy:
Devil's advocate, are these checks of any use? They go off semi-regularly, and no one seems to care when they do, so are they adding anything? Or would 'check there's at least one item' be enough?

andygout:
I think they have been ignored recently because they have been alerting falsely (and creating a lot of noise) and so are treated as boy who cried wolf, but if they are fixed then they will point to actual issues which people can see need addressing.

ironsidevsquincy:
They've never been acknowledged, even when they first started to go off
If you think there's value in keeping them, and they will be looked into, obviously go for it. BUT ON YOUR HEAD BE IT

PR #1416, opened 2016-12-19
1306 better api

### LINKED ISSUE #1306: Invalid Top Stories query to `next-api` to throw meaningful error
Currently if any part of the Top Stories query is invalid, an error is logged to Splunk here: [`server/data/bucket.js`](https://github.com/Financial-Times/next-front-page/blob/master/server/data/bucket.js#L46).

However, the actual cause of page failure is `"TypeError: Cannot read property 'items' of undefined"` (`data['topStory'].items[0]`) here: [`server/sections/topstories-hero-slice-1/model.js`](https://github.com/Financial-Times/next-front-page/blob/master/server/sections/topstories-hero-slice-1/model.js#L70).

The error should be caught at time of query error so as to provide a more meaningful error message that will help more with diagnosis.

aintgoin2goa:
Fixes #1306 
@andygout 

After much consideration I decided to log out only the errors from Next-API, not the query itself - I thought the output from the query could just get in the way, but I may be wrong,


andygout:
Am trying to trigger the `(err && err.name && err.name === 'HttpError')` conditional locally by making an intentionally invalid query, but the error name property is `Error` (rather than `HttpError`), but might have missed something.

As long as the Next API error message point us to the exact query then reckon we can dispense with the original query, as you suggest.

server/data/bucket.js
```diff
@@ -44,7 +44,21 @@ class DataBucket extends EventEmitter {
...
+		if(err && err.name && err.name === 'HttpError'){
+			err.responseBody().then(details => {
+				const errors = {};
+				if(details.errors && details.errors.forEach){
+					details.errors.forEach((error, index) => {
+						errors[`message_${index}`] = error.message;
+					});
+				}
+
+				log.error({event:'DATASOURCE_ERROR', bucket:this.name, host: err.url.hostname, url:err.url.pathname}, errors);
```
aintgoin2goa: fair point re: name and spacing, but splunk doesn't deal with nesting, names and values can just be primitives.  Have added an `errors` property which it all the error messages concatenated with a ","

aintgoin2goa:
@andygout  You need latest ft-poller for this to work

andygout:
Ah yup, can see it working now.

Are you not going to include changes discussed above?

PR #1417, opened 2016-12-20
Update video carousel and player

rowanbeentje:
Bit of a beefy upgrade which has been worked on in bits and bobs, I'm afraid; all fairly self-contained though!

- Sources video carousel from editorsPicks graphql query
- Updates standard view to a customised n-teaser
- Update the design to match the latest video/"inverted hero" mockups
- Alter core behaviour to display a overflow: scroll carousel, linking through to the video pages (which don't actually work in core yet, but should soon)
- Update carousel appearance to make better use of real estate on large screens, and increase contrast
- Fix display on various awkward mobile sizes
- Add video share support
- Improve accessibility a little


client/components/video-scroller/_video-scroller.scss
```diff
@@ -130,101 +134,184 @@ $video-scroller-transition: 150ms;
...
+			}
 		}
+	}
 
-		&:nth-child(3n+3) {
-			margin-left: ($stacked-width + $video-scroller-spacing) * -1;
+	.o-teaser__image-container {
+		background: oColorsGetPaletteColor('claret');
+		min-height: 110px;
+		img {
```
ironsidevsquincy: possible to get classes on these `img` elements?

client/components/video-scroller/_video-scroller.scss
```diff
@@ -130,101 +134,184 @@ $video-scroller-transition: 150ms;
...
+			transition: opacity 0.25s;
+			will-change: opacity;
```
ironsidevsquincy: woah!

client/components/video-scroller/_video-scroller.scss
```diff
@@ -130,101 +134,184 @@ $video-scroller-transition: 150ms;
...
+		content: '';
+		display: block;
+		position: absolute;
+		top: 0;
+		left: 0;
+		min-height: 60%;
+		right: $video-scroller-spacing;
+		z-index: 2;
+		pointer-events: none;
+		background: linear-gradient(rgba(0, 0, 0, 0.75), transparent);
```
ironsidevsquincy: how does this look on device that don't understand `linear-gradient`s?

client/components/video-scroller/_video-scroller.scss
rowanbeentje: That would be very useful - I'm also not a fan of the nth-child queries.  I think it was done that way originally to avoid extra class additions in the list but think that's worth overriding; I'll add some server-side class generation for the view to spit out.  (The `img` elements still won't have a class because they're within the n-teaser but at least the selector will be much simpler)

client/components/video-scroller/_video-scroller.scss
rowanbeentje: Great point - illegible white text on images.  Adding some solid colour + translucent colour fallbacks...

client/components/video-scroller/_video-scroller.scss
rowanbeentje: Flicker begone!

client/components/video-scroller/_video-scroller.scss
rowanbeentje: `img` classes used (oops).

I also moved the nth-child selectors to presenter-added classes instead but I'm having second thoughts about that.  Any feelings on keeping or chucking https://github.com/Financial-Times/next-front-page/pull/1417/commits/22790b5182542acc392d4ccb80544920e1c95c2c ?

ironsidevsquincy: Hmmm, I'm _slightly_ more in the keep it camp, as with those class names it's easier to understand what kind of element the css is styling

PR #1418, opened 2016-12-21
Fix: remove standalone video template

andygnewman:
cc: @leggsimon 
Now that video's use a standard teaser.


leggsimon:
Does this mean n-card can be fully deprecated now? (If it isn't already?)

PR #1427, opened 2017-01-05
ssh-safe custom bower entry

wheresrhys:
@i-like-robots This format is more forgiving of any quirks in a user's git setup


i-like-robots:
👍 

I want to kill this fork so badly but there we are

PR #1436, opened 2017-01-15
remove H level

lc512k:
This removes the H level from `n-teaser` title while preserving its style. 

Having `n-teaser` titles be `H`s is confusing for screen reader users, whose assistive tech strongly relies on HTML semantics. These H3/H2 elements end up being interpreted as subsection of whatever parent they're in, when they're really not.

cc @andygnewman @ironsidevsquincy @i-like-robots @keirog @leggsimon


lc512k:
Related to
Financial-Times/next-article#1767
https://github.com/Financial-Times/next-offline/pull/46

PR #1438, opened 2017-01-17
remove h3s from fastft headlines

lc512k:
@keirog I'm starting to think H3s are breeding when I'm not looking \o/


keirog:
Ha! I’ll keep an eye out for them…

🐇 
🐇 🐇 🐇

PR #1447, opened 2017-01-23
Video UI tweaks for keyboard users

keirog:
- Disables focusability of non-player links while the player is open
- Re-focuses the user back where they were after closing the video player

@lc512k


lc512k:
Nice one, @keirog !!

PR #1450, opened 2017-01-24
Remove extraneous TeaserTopStory queries

adgad:
Regional news and top stories (the non-top-story-one) shouldn't need to make the additional TeaserTopStory requests to Next API

@andygnewman @pixelandpage @ironsidevsquincy


andygnewman:
LGTM

PR #1452, opened 2017-01-25
More lenient missing items healthchecks

adgad:
@bjfletcher @ironsidevsquincy @pixelandpage

As discussed yesterday - I think it's still good to have some alerts for these, as if we're missing more than a few items it's probably a sign of something bigger wrong (but we can review that later).


ironsidevsquincy:
If the assumption is that if that many are missing than it's an 'us' problem, than 👍

adgad:
Yup, that's the assumption :)

PR #1465, opened 2017-02-02
Use CSS for video scroller position

ironsidevsquincy:
Simplifies it a tad


ironsidevsquincy:
Front page video carousel is going to get a re-design/write at some point, so we'll put up with any small visual issues for now

PR #1470, opened 2017-02-07
feat: show the compactview welcome tooltip 3 times

TheoLeanse:
screen shot 2017-02-07 at 15 48 19

show this lil fella 3 times, unless explicitly dismissed

- [x] show three times
- [x] unless dismissed


TheoLeanse:
wondering how if this will cause any problems https://github.com/Financial-Times/next-front-page/blob/master/client/main.js#L31-L35

PR #1473, opened 2017-02-09
Fastly soft purge

bjfletcher:
This PR soft-purges the frontpage on Fastly when there is data change.

Example log output (note: the initial burst of data changes is because app has just started up, and it debounces for 10 seconds):

```
event=DATA_CHANGED name=fastFt level=info
event=DATA_CHANGED name=internationalTopStories level=info
event=DATA_CHANGED name=ukTopStories level=info
event=DATA_CHANGED name=ukOtherHomePage level=info
event=DATA_CHANGED name=internationalOtherHomePage level=info
event=DATA_CHANGED name=regionalNews level=info
event=FASTLY_SOFT_PURGE_SUCCESS level=info
event=DATA_CHANGED name=regionalNews level=info
event=FASTLY_SOFT_PURGE_SUCCESS level=info
event=DATA_CHANGED name=regionalNews level=info
event=FASTLY_SOFT_PURGE_SUCCESS level=info
event=DATA_CHANGED name=regionalNews level=info
event=FASTLY_SOFT_PURGE_SUCCESS level=info
event=DATA_CHANGED name=fastFt level=info
event=FASTLY_SOFT_PURGE_SUCCESS level=info
event=DATA_CHANGED name=regionalNews level=info
event=FASTLY_SOFT_PURGE_SUCCESS level=info
event=DATA_CHANGED name=ukTopStories level=info
event=FASTLY_SOFT_PURGE_SUCCESS level=info

### CODE REVIEW

File: server/data/data-store.js [hunk 1]
```diff
@@ -12,10 +13,24 @@ class DataStore extends EventEmitter {
...
 			this.buckets.set(bucketConfig.key, bucket);
 		}
 	}
 
+	checkPurge (name){
+		logger.info({ event: 'DATA_CHANGED', name })
+		// debounce the soft-purging because we get a burst of data changes every ~1 minute
+		const later = () => {
+			this.timeout = null;
+			fetch('https://falcon-test-3.ft.com', { method: 'PURGE', headers: { 'Fastly-Soft-Purge': 1 } })
```
wheresrhys: needs to be www.ft.com

wheresrhys: I would attempt the purge 2 or 3 times - sometimes it will respond ok but the purge doesn't actually happen.

server/data/data-store.js
```diff
@@ -12,10 +13,24 @@ class DataStore extends EventEmitter {
...
+				.then(() => logger.info({ event: 'FASTLY_SOFT_PURGE_SUCCESS' }))
+				.catch(err => logger.err({ event: 'FASTLY_SOFT_PURGE_FAIL' }, err));
```
wheresrhys: I would add some metrics around failures, and we need a healthcheck for it too

server/controllers/front-page.js
```diff
@@ -60,8 +60,8 @@ function render (req, res, edition){
 	};
 
 	setupMarketsHeader(res.locals.flags, renderParams);
-	res.set('Surrogate-Control', 'max-age=30, stale-while-revalidate=60, stale-if-error=86400');
-	res.set('Cache-Control', res.FT_NO_CACHE);
+	res.set('Surrogate-Control', 'max-age=3700, stale-while-revalidate=60, stale-if-error=86400');
```
wheresrhys: I'd set up the mechanism, monitoring first, check everything seems to function ok, and then after that try out raising the cache max-age

server/data/bucket.js
```diff
@@ -70,6 +70,9 @@ class DataBucket extends EventEmitter {
 
 	store (data){
 		if(this.storage === DataBucket.storage.MEMORY){
+			if (JSON.stringify(this._data) !== JSON.stringify(data)) {
```
wheresrhys: worth using a tool such as `deep-diff`? Is graphql deterministic as to property order?

server/controllers/front-page.js
wheresrhys: ah, just seen your note in slack about trying it in a branch first so I guess you need this for that.

server/data/bucket.js
ironsidevsquincy: Front what I've seen it is, but I don't know that for a fact, so probably best to go with a lib to determine

bjfletcher:
@wheresrhys just to reassure you, this isn't going into `master` anytime soon - not without some thought put into monitoring, recovery, fault tolerance, etc., for this.  I'm more interested to see if this works and fixes the `stale-if-error` problem in the testing environment.  If it works, fantastic!! Then we'll look at how to make this into something we can use with the production/live system. 🎉

PR #1484, opened 2017-02-16
use request instead of node-fetch as the latter does a hard purge :/

bjfletcher:
Fastly and I have discovered that the soft-purge calls were actually hard-purge calls. Tried everything with `node-fetch` but could not, then tried `request` which worked. So, switching to this package for now.


server/data/data-store.js
```diff
@@ -50,13 +50,16 @@ class DataStore extends EventEmitter {
...
 		};
-		fetch(url, { method: 'POST', headers })
-			.then(fetchres.json)
-			.then(() => logger.info({ event: 'SOFT_PURGE_SUCCESS', url }))
-			.catch(err => logger.error({ event: 'SOFT_PURGE_FAIL', url }, err));
+		request({ url, method: 'POST', headers }, (err, resp, body) => {
+			if (!err && resp.statusCode === 200) {
+				logger.info({ event: 'SOFT_PURGE_SUCCESS', url, body });
+			} else {
+				logger.error({ event: 'SOFT_PURGE_FAIL', url, err, body });
```
ironsidevsquincy: `err` should be its own argument

server/data/data-store.js
```diff
@@ -50,13 +50,16 @@ class DataStore extends EventEmitter {
 		const url = 'https://api.fastly.com/service/133g5BGAc00Hv4v8t0dMry/purge/frontpage';
 		const headers = {
 			'Fastly-Key': process.env.FASTLY_KEY,
-			'Fastly-Soft-Purge': 1,
+			'Fastly-Soft-Purge': '1',
```
ironsidevsquincy: was this why `fetch` didn't work, because the header value needed to be a string?

PR #1491, opened 2017-02-23
Update to n-ui v4

ironsidevsquincy:


client/main.scss
```diff
@@ -1,37 +1,30 @@
...
 	$o-labels-is-silent: false;
 	@import 'o-labels/main';
 	$o-teaser-is-silent: false;
 	@import 'o-teaser/main';
 	$o-teaser-collection-is-silent: false;
 	@import 'o-teaser-collection/main';
+
+	@import 'n-ui/tour-tip/main';
+
+	@import 'components/ads/ads';
```
wheresrhys: as the ads component handles its own criticalness, does it need to be within this block?

PR #1500, opened 2017-03-02
o-hoverable override

gvonkoss:
@onishiweb ?


onishiweb:
@gvonkoss ?

lc512k:
@onishiweb iirc it's the same issue we were looking at yesterday where `.o-hoverable-on a:hover` is overriding the hover style in o-teaser. Could we increase specificity there rather than `!important`?

onishiweb:
If needed I can add the `.o-hoverable-on` class to the specificity of the selector in `o-teaser`?

gvonkoss:
sounds like that would be better than `!important` 🙈

onishiweb:
Released: https://github.com/Financial-Times/o-teaser/releases/tag/v1.11.0 see if that works :)

PR #1506, opened 2017-03-09
Dirty fix for the compact view’s welcome panel

keirog:
It turns out there’s a version of the welcome panel that advertises the compact view (as well as the version for when compact view is already on). When we turned off the `welcomePanel` flag recently, the compact _advert_ version broke. This naff change fixes it and avoids yak shaving. Peace and love ✌️ 🕊


keirog:
Relies on https://github.com/Financial-Times/n-ui/pull/869

